PR #3, opened 2015-02-01
Nim improvements

def-:
- Always compile with clang backend, as it often has better performance
- Clean up code a bit
- Minor speedups

Julia's matmul benchmark looks strange. Is it really that much faster than everything else? Do they have some magic?

### DIFF
--- a/base64/build.sh
+++ b/base64/build.sh
@@ -5,5 +5,5 @@ gcc -O3 -std=c99 -o base64_c test.c
 scalac -optimize test.scala
 rustc --opt-level 3 test.rs -o base64_rs
 dmd -ofbase64_d -O -release test.d
-nim c -o:base64_nim -d:release --verbosity:0 --hints:off test.nim
+nim c -o:base64_nim -d:release --cc:clang --verbosity:0 test.nim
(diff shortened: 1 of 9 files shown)


kostya:
thanks, i think need to revert matmul changes, because all other implementations use double arrays, so its not fair. 

julia native matmul https://github.com/kostya/benchmarks/blob/master/matmul/matmul-native.jl faster,  because they implement self algorithm, its not fair to compare with others, need to compare with julia https://github.com/kostya/benchmarks/blob/master/matmul/matmul.jl

def-:
Julia uses float32 as well, right?

kostya:
yes but even with float32 its far from top

kostya:
funny that brainfuck clang version 35% faster, but others not so much.

def-:
Yes, it really depends a lot on the code. Sometimes GCC even ends up faster. Same with C and C++.

MartinNowak:
> funny that brainfuck clang version 35% faster, but others not so much.

Nim is so fast on brainfuck because it has the fastest hash table implementation for small values (open addressing, storing values in head). The difference between clang and gcc is quite drastic though.

PR #7, opened 2015-03-31
avoid a few allocations in havlok benchmark

MartinNowak:


kostya:
thanks, i will try it tomorrow

MartinNowak:
Thank you, those are really good benchmarks.
They hit quite a few pain points in D's current implementation though.
http://forum.dlang.org/post/mff3n4$gnm$1@digitalmars.com

PR #8, opened 2015-03-31
update D compilers

MartinNowak:
- use 2.067.0 with much better GC performance
- add LDC v0.15.1 with old GC, but faster codegen

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,8 @@ Intel(R) Core(TM) i5-2400 CPU @ 3.10GHz (Ubuntu 14.04.1 LTS x86_64)
 * Nim Compiler Version 0.10.2 (2014-12-29) [Linux: amd64]
 * `Crystal 0.5.5 [d814c6c] (Fri Dec 12 22:50:10 UTC 2014)`
 * go version go1.4 linux/amd64
-* DMD64 D Compiler v2.066.1
+* DMD64 D Compiler v2.067.0
+* LDC D Compiler v0.15.1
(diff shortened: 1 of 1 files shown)


MartinNowak:
See http://dlang.org/changelog.html#new2_067.0 and https://github.com/ldc-developers/ldc/releases/tag/v0.15.1.

kostya:
i trying gdc, but json failed with:
```
gdc -o json_d_gdc -O3 -frelease -finline test.d
test.d:7: error: cannot implicitly convert expression (parseJSON(text, -1).object) of type JSONValue[string] to JSONValue
```

kostya:
ldc2 -ofjson_d_ldc -O5 -release -inline json.d
```
Error: module json is in file 'json.d' which cannot be read
import path[0] = /opt/ldc2-0.15.2-beta1-linux-x86_64/bin/../import/ldc
import path[1] = /opt/ldc2-0.15.2-beta1-linux-x86_64/bin/../import
```

kostya:
added ldc, gdc

MartinNowak:
It's a bit unfortunate, they use different command lines. You can use `ldmd2` (ships with ldc) and [`gdmd`](https://github.com/D-Programming-GDC/GDMD) as wrappers or adopt the command line invokation.

MartinNowak:
I'll look at the failures when I get home.

MartinNowak:
Thanks for adding them, see #10 for LDC and json.

PR #12, opened 2015-04-07
Julia improvements

tkelman:
Use dispatch, not member functions. Julia types are not like conventional OO classes. With these changes the performance is more in line with the (AOT-)compiled languages.

I didn't touch the others since they're mostly measuring the the load time of the JSON package, or demonstrating clearly why it's a good idea to use a fast BLAS for linear algebra.

### DIFF
--- a/base64/test.jl
+++ b/base64/test.jl
@@ -11,7 +11,7 @@ function main()
   t = time()
   s = 0
   for i in range(0, tries)
-    str2 = bytestring(encode(Base64, str))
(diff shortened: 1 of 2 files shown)


kostya:
thanks, i will try it little later

IainNZ:
@kostya did you learn that style of writing Julia from a tutorial? If so, do have the URL for it?

kostya:
i don't remember, seems i find some examples on github.

kostya:
brainfuck example 35 times faster )

PR #13, opened 2015-04-09
replace readText with read

MartinNowak:
- readText performs additional UTF validation


kostya:
thanks, i will update time later

PR #15, opened 2015-04-22
Update brainfuck.d

schancel:
Optimization of DLang benchmark.

### DIFF
--- a/brainfuck/brainfuck.d
+++ b/brainfuck/brainfuck.d
@@ -2,14 +2,16 @@ import std.algorithm;
 import std.stdio;
 import std.file;
 import std.array;
+import std.conv;
 
-class Tape {
+struct Tape {
   int pos;
   int[] tape;
 
-  this() {
-    pos = 0;
-    tape ~= 0;
+  @disable this();
+
+  this(int[] t) {
+    tape = t;
   }
 
 final:
@@ -45,10 +47,11 @@ class Program {
       pc++;
       code ~= c;
     }
+    bracket_map.rehash;
   }
 
   void run() {
-    Tape tape = new Tape();
+    auto tape = Tape(new int[100]);
     for (int pc = 0; pc < code.length; pc++) {
       switch (code[pc]) {
         case '+':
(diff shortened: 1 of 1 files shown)


kostya:
other implementations used bracket_map as language hash map, so change this is not fair. Also, not used anywhere some reserving. And seems switch from class to struct not changing performance.

brainfuck/brainfuck.d
```diff
@@ -22,11 +24,14 @@ final:
 
 class Program {
   string code;
```
Orvid: In the constructor, allocate this as a string of the same length as the input text, and then assign to each index. At the end, slice it down to the used length, that gets rid of need for the code to be appended character by character to the string.

Orvid:
The class->struct change isn't a performance change, but rather, is a change to bring it into line with how D code is normally written. It also means that no type info will be emitted for the type, as it's not needed.

schancel:
Actually, changing it to a struct causes it to be a stack variable and avoid a double deref for each lookup of the tape.

Compare, struct:
```./brainfuck mandel.b  99.92s user 0.20s system 99% cpu 1:40.14 total```
class: 
```./brainfuck mandel.b  131.48s user 0.30s system 99% cpu 2:11.82 total```

File: brainfuck/brainfuck.d [hunk 2]
```diff
@@ -45,10 +47,11 @@ class Program {
       pc++;
       code ~= c;
     }
+    bracket_map.rehash;
   }
 
   void run() {
-    Tape tape = new Tape();
+    auto tape = Tape(new int[100]);
```
schancel: I know you had concerns around this line, but the amount of tape used by these tests is around 8 bytes.   Pre-allocating a small region like this doesn't change the benchmarks.  It's only to make sure that the tape is initialized properly since structs do not allow default constructors to be specified in DLang.

brainfuck/brainfuck.d
```diff
@@ -70,7 +73,7 @@ class Program {
           if (tape.get() != 0) pc = bracket_map[pc];
           break;
         case '.':
-          printf("%c", tape.get());
+          write(tape.get().to!char);
```
schancel: Printf is not type safe in D and should not be used.

kostya:
i trying it by parts, changing to struct seems not give any speedup, somewhere it slower:
https://github.com/kostya/benchmarks/commit/7e296a229fe771ea40832ec7da3c5cae6458ad57

PR #16, opened 2015-04-23
New brainfuck d

schancel:
Reworking of the D example.  Is this an acceptable implementation?


kostya:
please don't change the algorithm :) all implementations using the same.

schancel:
Okay, well the other PR doesn't change the algo and does result in a good speedup.

PR #17, opened 2015-04-25
Monotests

chalucha:
Added some simple benchmarks for mono to compare with others.


kostya:
thanks, how about brainfuck?

PR #18, opened 2015-04-25
Added mono version of brainfuck

chalucha:
Here you go with the brainfuck C# version.
Will look to havlak too, but its a bit longer.


kostya:
thanks, i changed struct to class, because of error `brainfuck.cs(14,16): error CS0568: Structs cannot contain explicit parameterless constructors`
Havlak it too hard, not torturing yourself, and i not think it is a good bench.

PR #19, opened 2015-04-25
Added C# variant of havlak benchmark

chalucha:
So here it is.. ;)
I did not try to figure what its doing, but just rewritten the D variant into C#. They are pretty close with syntax, only some differences with maps, dynamic arrays, foreach, and some minor things.


kostya:
thanks, i will update it later

PR #23, opened 2015-04-26
Dlang

chalucha:
Make classes final, and add some attributes to methods definitions to speed up slightly

### DIFF
--- a/havlak/havlak.d
+++ b/havlak/havlak.d
@@ -151,15 +151,15 @@ class HavlakLoopFinder {
   CFG cfg;
   LSG lsg;
 
-  static int UNVISITED = -1;
-  static int BB_TOP = 0;
-  static int BB_NONHEADER = 1;
-  static int BB_REDUCIBLE = 2;
-  static int BB_SELF = 3;
-  static int BB_IRREDUCIBLE = 4;
-  static int BB_DEAD = 5;
-  static int BB_LAST = 6;
-  static int MAXNONBACKPREDS = (32 * 1024);
+  enum UNVISITED = -1;
+  enum BB_TOP = 0;
+  enum BB_NONHEADER = 1;
+  enum BB_REDUCIBLE = 2;
+  enum BB_SELF = 3;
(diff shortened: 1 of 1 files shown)


havlak/havlak.d
```diff
@@ -356,64 +364,67 @@ int findHavlakLoops(CFG cfg)
...
+        BasicBlockEdge(cfg, bb0 + 1, bb0 + 3);
+        BasicBlockEdge(cfg, bb0 + 2, bb0 + 3);
+        return bb0 + 3;
+    }
 
-  void buildConnect(int _start, int _end) {
-    new BasicBlockEdge(cfg, _start, _end);
-  }
+    void buildConnect(int _start, int _end) {
+        BasicBlockEdge(cfg, _start, _end);
```
Temtaime: ???

chalucha: Just an indentation caused by:
+  @safe pure
+  {

removed unneeded new because BasicBlockEdge is a struct

kostya:
oh, so many hacks, i dont like hacks, which make bad code readability, d not changed, ldc used 25s
also gdc fails to compile: `havlak.d:35: error: safe function 'havlak.CFG.createNode' cannot call system function 'object.AssociativeArray!(int, BasicBlock).AssociativeArray.get'
havlak.d:47: error: safe function 'havlak.CFG.getNumNodes' cannot call system function 'object.AssociativeArray!(int, BasicBlock).AssociativeArray.length'
`

chalucha:
So what about this?
Just added module level final to make all classes final (unfortunately not set as default in D), which has a "major" impact for results over the rest previously used attributes as @safe, pure, etc.
Replaced printf with writeln (no perf hit)
Changed constants to enums (no perf hit)
Removed new from struct initialization - not needed.

kostya:
thanks, but what this final means?

chalucha:
It means that final classes cannot be subclassed. In D all methods of class are virtual by default and being so means some ovehead calling them.
I dont understand this decision, but it is how it is ;-)

PR #26, opened 2015-06-06
Use BTreeMap instead of HashMap in brainfuck.rs

steveklabnik:
HashMap uses a cryptographically sound hashing algorithm, and so is much slower than BTreeMap. On my system:

```
$ bash run.sh
Rust
ZYXWVUTSRQPONMLKJIHGFEDCBA
8.53s, 2.1Mb

$ bash run.sh
Rust
ZYXWVUTSRQPONMLKJIHGFEDCBA
4.30s, 2.2Mb
```

### DIFF
--- a/brainfuck/brainfuck.rs
+++ b/brainfuck/brainfuck.rs
@@ -4,7 +4,7 @@ use std::io::prelude::*;
 use std::vec::Vec;
 use std::io;
 use std::env;
-use std::collections::HashMap;
+use std::collections::BTreeMap;
 
 struct Tape {
   pos: usize,
@@ -23,13 +23,13 @@ impl Tape {
 
 struct Program {
   code: Vec<char>,
-  bracket_map: HashMap<usize, usize>
+  bracket_map: BTreeMap<usize, usize>
 }
 
 impl Program {
   fn new(content: String) -> Program {
     let mut code: Vec<char> = Vec::new();
-    let mut bracket_map = HashMap::new();
+    let mut bracket_map = BTreeMap::new();
     let mut leftstack = Vec::new();
     let mut pc = 0;
 


tilpner:
On my system, with BTreeMap:

    Rust
    ZYXWVUTSRQPONMLKJIHGFEDCBA
    5.96s, 6.2Mb

with VecMap (requires Nightly)

    Rust
    ZYXWVUTSRQPONMLKJIHGFEDCBA
    3.00s, 6.1Mb

This just shows how much the performance is affected by the used Map implementation.

filsmick:
@tilpner I wonder, why does the BTreeMap implementation use 6.2 Mb on your system, but 2.2 on @steveklabnik's?

tilpner:
@filsmick Yes, I wondered that too. I've got no idea. I made the same changes as @steveklabnik...

tilpner:
@filsmick Due to the great amount of architecture-dependent integers in the benchmark... a difference of 32 bit vs 64 bit might have caused this.

filsmick:
@tilpner Good catch, that's good to know.

eddyb:
`VecMap` uses twice as much space as you actually need, so you could switch to a `Vec<u32>` and use `u32::MAX` as the "hole" value (we should do this in the future with `VecMap` and `Default` or a custom trait, to indicate the "hole" value).

eddyb:
Also, keep in mind `std::map` in C++ is *not* a hash map, it's an ordered (likely tree-based) map.
Only in C++11 was `std::unordered_map` added, which *is* a hash map.
Using `HashMap` in Rust instead of `BTreeMap` is an unfair comparison, no matter the choice of the hash algorithm.

kostya:
thanks, i ll update it little later

kostya:
btw, why rust not have default `Map` (which is just alias for BTreeMap), like `Vec`. When i searching for map in rust docs, first what i find was HashMap and i just use it. Other languages give default map, like D: `int[int] bracket_map;`

filsmick:
+1, a default map would help minimize beginner confusion, as BTreeMap seems to be the default choice.

eddyb:
@kostya because it's up to the user to choose [the right tool for the job](http://doc.rust-lang.org/std/collections/).
`BTreeMap` is actually never used in the compiler, which employs `HashMap` and `HashSet` with a custom FNV hasher.

kbknapp:
In addition to switching to `VecMap` making these very small changes as well gets it down to `2.10s, 6.1Mb` on my machine (from the original `11.09s, 6.1Mb` on my machine, for comparison)

- [ ] Change `env::args()` to `env::args_os()`
- [ ] Use `String::with_capacity()` (and then `s.shrink_to_fit()`) instead of `String::new()`
- [ ] Finally change the `tape` field to a `Vec<u8>` (as well as the corresponding returns and such)

kostya:
VecMap is unfair here, because need to change in other bf implementations map to vector too, and bf program can be super large (vecmap would super inefficient by memory here).

vldm:
`VecMap` is a map optimized for small integer keys. You should define what you benchmarking. If you want to benchmark quality of code generation, then you should use same algorithm\structure in all language. In this case you should rewrite nim's `Table` to `Tree` like in other languages. Or you can test quality of the standard library, then u should use all possible container keeping your basic algorithm unchanged. In all case `VecMap` is fair as long as table fair in other languages.

Veedrac:
I agree with kostya: using `VecMap` would make the implementations incomparable. Same with using `Vec<u8>` which isn't what other implementations are doing.

ssokolow:
@Veedrac In that case, definitely best to take the following steps so the benchmark isn't misleading:

1. Amend the README to make it clear that code generation is what is being tested and, as such, the implementations may ignore faster data structures which are only present in the standard libraries of some of the languages under comparison.

2. Rewrite nim's `Table` to `Tree` so it's an apples-to-apples comparison.

kostya:
changing to Vec<u8> make it little slower. String::new() and env::args_os() not affected to benchmark.

PR #29, opened 2015-06-06
Improve Rust matmul benchmark

Manishearth:
On my i7, this goes from

4.13s, 102.2Mb

to

3.87s, 78.2Mb

The `with_capacity()` preallocates since we already know the final length.

A more ergonomic and common way to do it would be to do:

```rust
fn newmat(x: usize, y: usize) -> Vec<Vec<f64>> {
  let inner = vec![0; y];
  vec![inner; x];
}
```

This has the same perf, but looks cleaner. It's also what most Rust folks would do if asked to "make a 2D vector of x and y dimensions".

I can switch my PR to do this instead if you'd like.

### DIFF
--- a/matmul/matmul.rs
+++ b/matmul/matmul.rs
@@ -1,9 +1,9 @@
 use std::env;
 
 fn newmat(x: usize, y: usize) -> Vec<Vec<f64>> {
-  let mut r = Vec::new();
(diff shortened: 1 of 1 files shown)


kostya:
https://github.com/kostya/benchmarks/commit/048d4dbdea84f0a16bfa09175bb7dbcff5369342

PR #30, opened 2015-06-06
Improve Rust base64 implementation

cristicbz:
Two things really:
 * Don't build the initial string up in a for loop, causing many many allocations.
 * Enable link time optimization.

I still kept left the implementation reallocating the string every iteration (seems fair since that's what C++ also does).

Please don't rerun the benchmarks until https://github.com/rust-lang/rustc-serialize/pull/118 is also merged . I'll make some noise on the PR when that happens, thanks!

### DIFF
--- a/base64/base64.rs/Cargo.toml
+++ b/base64/base64.rs/Cargo.toml
@@ -6,3 +6,5 @@ version = "0.0.1"
 rustc-serialize = "*"
 time = "*"
 
+[profile.release]
+lto = true
--- a/base64/base64.rs/src/bin/base64.rs
+++ b/base64/base64.rs/src/bin/base64.rs
@@ -3,30 +3,26 @@ use rustc_serialize::base64::{ToBase64, FromBase64, STANDARD};
 extern crate time;
 use time::precise_time_ns;
 
+const STR_SIZE: usize = 10000000;
+const TRIES: usize = 100;
+
 fn main() {
-  let str_size = 10000000;
-  let tries = 100;
+  let input = vec![b'a'; STR_SIZE];
+  let mut output = String::new();
 
-  let mut str: String = "".to_string();
-  let mut str2: String = "".to_string();
-  for _ in 0..str_size { str.push_str("a"); }
-  let bytes = str.as_bytes();
-  
-  let mut t = precise_time_ns();
-  let mut s = 0;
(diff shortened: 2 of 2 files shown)


cristicbz:
The pull request was merged, feel free to merge at your earliest convenience, thanks!

semmaz:
Well, to be fair, this PR (and current rust implementation) [encodes](http://doc.rust-lang.org/num/rustc_serialize/base64/trait.ToBase64.html) to `String` from `Vec<u8>` and [decodes](http://doc.rust-lang.org/num/rustc_serialize/base64/trait.FromBase64.html) to `Vec<u8>` from `String`. Think I'll make alternative PR with `String`'s as input and output.

cristicbz:
The input in the original code is a slice of bytes - doesn't matter if it comes from a Vec or a String. If you want you can initialize the input with String::with_capacity(SIZE) and then extend with repeat('A').take(SIZE). It won't make any perf difference, I used Vec to bring it more in line with the C++ impl.

semmaz:
True, it doesn't matter if input is Vec or a String. It does matter for output though, from 1.48s for slice of bytes to 2.02 for String on my system. Try it yourself, change this [line](https://github.com/cristicbz/benchmarks/commit/af6de15ad81cec2b19e8e237e3a0e93e0dec5f79#diff-2de25544588305be691775e62e238aa1R24) to:
```rust
    sum += String::from_utf8(output.from_base64().unwrap()).unwrap().len();
```

semmaz:
I assume this happens because "String::from_utf8" use additional checking for valid UTF-8 bytes.
If I'll use unsafe version:
```rust
    sum += unsafe { String::from_utf8_unchecked(output.from_base64().unwrap()).len() }
```
time will go down to 1.48s for me again.

cristicbz:
Yes, that's why, but the original code didn't do that, nor does the c++ impl. Also I think when people encode things in base 64, it's a binary blob - not much point base64 encoding a utf8 string (bar url safe stuff, but even then I think it's more common to escape chars in human readable strings rather than use b64)

kostya:
thanks, nice improved base64 https://github.com/kostya/benchmarks/commit/048d4dbdea84f0a16bfa09175bb7dbcff5369342

PR #32, opened 2015-06-06
Cleaned up json test a bit.

itsmontoya:


json/test.go
```diff
@@ -30,5 +37,7 @@ func main() {
   }
 
   len := float64(len(jobj.Coordinates))
+  fmt.Printf("%.8f\n%.8f\n%.8f\n", x/len, y/len, z/len)
+}
   fmt.Printf("%.8f\n%.8f\n%.8f\n", x / len, y / len, z / len)
```
kostya: why double print here?

PR #33, opened 2015-06-06
Optimizations for brainfuck/C++; 6.87sec -> 2.90sec on i3-2100T

anttirt:
You can use -DUSE_FOO=0 for any of the four specific optimizations to see how they impact performance.

Tested with g++ (Ubuntu 4.8.2-19ubuntu1) 4.8.2

### DIFF
--- a/brainfuck/brainfuck.cpp
+++ b/brainfuck/brainfuck.cpp
@@ -3,6 +3,22 @@
 #include <map>
 #include <fstream>
 
+#ifndef USE_CODE_VECTOR
+#define USE_CODE_VECTOR 1
+#endif
+
+#ifndef USE_FLAT_MAP
+#define USE_FLAT_MAP 1
+#endif
+
+#ifndef USE_CONST_SIZE
+#define USE_CONST_SIZE 1
+#endif
+
+#ifndef USE_COMPACT_OPCODES
+#define USE_COMPACT_OPCODES 1
+#endif
+
 using namespace std;
 
 class Tape {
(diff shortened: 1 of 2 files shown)


kostya:
you also change bracket_map to vector :_) 
all implementations used standard language map, so it unfair.
not sure if other changes have any meaning, gcc should optimize other cases by itself. and code is so dirty.

anttirt:
> so it is unfair.

First off, Nim's Table is a hash table; C++ std::map is a red-black tree, but std::unordered_map is also available, so that's already a significant difference in the libraries used. Secondly, I think the benchmark itself is kind of unfair and misleading because the benchmark is not comparing the language or compiler.

The benchmark compares how well the standard library's map/table is optimized for *extremely small maps*; there are only 27 entries in the map.

The other changes do have a small but clear effect as well. With all of them disabled, I get 6.45sec and with only -DUSE_FLAT_MAP=0 I get 6.01sec.

kostya:
i tested unordered_map it was even slower. in rust for example was hash map, but when move to btree it became faster https://github.com/kostya/benchmarks/pull/26.

i don't want to do super hacks to win benchmark. i want to implement algorithm idiomatically (vector is not good, because bf code can be super big), in the same way in different languages, and see, what performance they give. and may be apply little performance tricks.

yes, my way of benchmarking is not fair, but it is what i wanted to measure.

anttirt:
> yes, my way of benchmarking is not fair, but it is what i wanted to measure.

I see. I'd appreciate it if you at least added a comment to the root README.md noting that many of the benchmarks measure library performance and not language/compiler performance.

Manishearth:
> i don't want to do super hacks to win benchmark. i want to implement algorithm idiomatically

This isn't as cut and dry as it sounds. "Idiomatically" is context-sensitive. When there are multiple maps available, the more idiomatic map depends on what you're trying to do.

In Rust's case, for example, [this is explicitly stated in the docs](http://doc.rust-lang.org/std/collections/). When you say "I want a hashmap", that question isn't enough.

What this often does end up benchmarking is what naively written code is like :smile:

kostya:

PR #34, opened 2015-06-07
Update rust brainfuck with a much more rust-oriented approach

daboross:
This updates brainfuck.rs to use a more rust oriented approach to the rust brainfuck, using an enum to store instructions and loop positions rather than a `Vec<char>` and `HashMap`.

I've tried to keep to the coding style used in the previous file as much as possible while using a more rustic approach to the execution itself.

### DIFF
--- a/brainfuck/brainfuck.rs
+++ b/brainfuck/brainfuck.rs
@@ -1,10 +1,20 @@
 use std::fs::File;
-use std::path::Path;
+use std::path::PathBuf;
 use std::io::prelude::*;
 use std::vec::Vec;
 use std::io;
 use std::env;
-use std::collections::HashMap;
+
+pub enum Instruction {
+  MoveRight, // `>`
+  MoveLeft, // `<`
+  Increment, // `+`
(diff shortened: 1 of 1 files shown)


kostya:
why close? seems not bad, i can't accept it right now, because other implementations is different, but may be reimplement other tests like this, to remove questions about hash_map, tree_map.

daboross:
I just realized that it would probably be not great to have one program implement it completely differently from the others. I'll reopen if you want to work on changing the other programs to behave like this, although I wouldn't know how to do this in many other languages.

kbrock:
@kostya Are you testing the various languages using the same basic implementation?
Or are you testing idiomatic code?

May want to state your preference.

I did notice a number of multi arrays and thought a single array may be quicker. -- `a[x,y]` vs a[x*WIDTH+y]`, but don't know your preference on this matter.

Would be cheap to implement one way in one language and the other way for another language.

dbohdan:
@kbrock I second your question.

PR #36, opened 2015-06-08
Changed bracket_map to RedBlackTree, which is faster ATM

tchaloupka:
D's associative array is not fast enough yet. Can be replaced with RBTree structure, which is much faster.

Tuple is used as a key value pair here.
EqualRange method of rbTree returns all tuples with the same key, front is used to pick first (and only one).

This change gets it (LDC2) from 6s to 3.8s on my PC.

### DIFF
--- a/brainfuck/brainfuck.d
+++ b/brainfuck/brainfuck.d
@@ -22,8 +22,11 @@ final:
 };
 
 class Program {
+  import std.container.rbtree;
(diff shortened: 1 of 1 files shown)


kostya:
thanks, but i going to remove bracket_map at all, because many questions about this.

kostya:

PR #38, opened 2015-06-09
Rust: migrated json to use serde.

cristicbz:
rustc-serialize is pretty much a rustc internal crate. The state of the art serialization in the Rust world currently is serde. This PR migrates the Rust json benchmark to use serde. This cuts the benchmark time from about 5.5s to about 1.3s on my machine. 

Note: since serde is strict by default (i.e. unknown fields are treated as an error) the solution is to create a small dummy type to eat up missing fields; I hope you consider this in the spirit of the benchmark---everything's still validated, it's just not allocated or saved in the deserialized struct, similar to rustc-serialize's behaviour. 

Note 2: serde-macros only works on nightly currently. If this is a problem, let me know, I can use serde-codegen which is a bit more verbose, but runs on stable as well (by introducing an extra build pass).


json/json.rs/src/main.rs
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  println!("{}", x / len);
+  println!("{}", y / len);
+  println!("{}", z / len);
+}
+
+// A simple type which deserializes by silently consuming all its input, allowing a field to be
+// skipped.
+struct Skip;
```
kostya: thanks for pr, btw why this Skip struct can't no be implemented inside serde crate?

cristicbz: It's a bit bigger, I know. I think serde, for now, just wasn't designed with this use case in mind. I think a more general solution could be implemented in serde.

However, I think that having this additional snippet in the benchmark is nice to also show off how you could customize deserialization if you wanted to.

Vikaton:
This makes Rust slower than it was

PR #41, opened 2015-06-16
Optimization for brainfuck.scala (8.96s -> 6.28s on i5-4670)

orionll:


kostya:
thanks, but i going to remove bracket_map at all, because many questions about this.

kostya:

PR #43, opened 2015-07-01
Tweak C++ and Julia BF programs; add Felix and OOC

kirbyfan64:
- In Julia, explicit returns are considered bad style.
- In C++, you don't need to create a new string to append it to the other string.
- Felix is *fast*!


kostya:
hi, can you add julia, felix, and ooc implementations for new brainfuck implementation?
https://github.com/kostya/benchmarks#brainfuck-v20.

kirbyfan64:
@kostya Sure! Just give me a few days.

Issue #1, opened 2014-10-16
Json benchmark


asterite:
Hi!

It would be interesting to see a benchmark comparing json parsing. You can try with this big json: https://github.com/zeMirco/sf-city-lots-json

We tried to optimized json parsing in Crystal a lot and we believe it might be one of the fastest out there. And, as usual, it's implemented in Crystal itself.

Here's a code that you can try:

``` ruby
require "json"

text = File.read("citylots.json")
json = Json.parse(text) as Hash
puts json.length
```

Thanks!

PR #46, opened 2015-10-12
Added fast.json D library beating RapidJSON

mleise:
Since RapidJSON claims to be the fastest I thought I'll accept the challenge. For better comparability I chose the Gcc based backend.

### DIFF
--- a/json/.gitignore
+++ b/json/.gitignore
@@ -1,3 +1,4 @@
+fast
 rapidjson
 Newtonsoft.Json*
 gason
--- a/json/build.sh
+++ b/json/build.sh
@@ -16,6 +16,11 @@ go build -o json_go test.go
 gccgo -O3 -g -o json_go_gccgo test.go
 g++ -O3 test_boost.cpp -o json_boost_cpp
 
+if [ ! -d fast ]; then
+  git clone --depth 1 --branch v0.3.0 https://github.com/mleise/fast.git
+fi
+gdc -o json_d_fast -O3 -frelease -fno-bounds-check test_fast.d fast/source/fast/cstring.d fast/source/fast/buffer.d fast/source/fast/helpers.d fast/source/fast/json.d fast/source/fast/parsing.d fast/source/fast/intmath.d
+
 if [ ! -d rapidjson ]; then
   git clone --depth 1 https://github.com/miloyip/rapidjson.git
 fi
--- a/json/run.sh
+++ b/json/run.sh
@@ -12,10 +12,12 @@ echo Go
 ../xtime.rb ./json_go
 echo GccGo
 ../xtime.rb ./json_go_gccgo
-echo D
+echo D DMD
 ../xtime.rb ./json_d
 echo D Gdc
 ../xtime.rb ./json_d_gdc
+echo D Gdc fast
+../xtime.rb ./json_d_fast
 echo D Ldc
 ../xtime.rb ./json_d_ldc
 echo Nim Gcc
--- a/json/test_fast.d
+++ b/json/test_fast.d
@@ -1,0 +1,26 @@
+// By default fast.json validates only portions of JSON that
+// are read out as a compromise between speed and correctness.
+// That includes UTF-8 validation for file based JSON.
+
+import core.stdc.stdio;
+import fast.json;
+
+struct Coord { double x, y, z; }
+
+void main()
+{
+    double x = 0, y = 0, z = 0;
+
+    auto json = parseJSONFile("./1.json");
+    auto coords = json.coordinates.read!(Coord[]);
+
+    foreach (ref coord; coords)
+    {
+        x += coord.x;
+        y += coord.y;
+        z += coord.z;
+    }
+
+    auto len = coords.length;
+    printf("%.8f\n%.8f\n%.8f\n", x / len, y / len, z / len);
+}


kostya:
thanks, many compile errors:
```
fast/source/fast/buffer.d:54: error: user defined attributes must be first
fast/source/fast/buffer.d:62: error: user defined attributes must be first
fast/source/fast/buffer.d:62: error: no identifier for declarator property
fast/source/fast/buffer.d:62: error: semicolon expected, not 'inout'
fast/source/fast/buffer.d:69: error: user defined attributes must be first
fast/source/fast/buffer.d:69: error: semicolon expected, not 'capacity'
fast/source/fast/buffer.d:69: error: function declaration without return type. (Note that constructors are always named 'this')
fast/source/fast/buffer.d:70: error: no identifier for declarator const capacity()
fast/source/fast/buffer.d:118: error: user defined attributes must be first
fast/source/fast/buffer.d:118: error: no identifier for declarator property
fast/source/fast/buffer.d:118: error: semicolon expected, not 'inout'
fast/source/fast/buffer.d:125: error: user defined attributes must be first
fast/source/fast/buffer.d:125: error: semicolon expected, not 'length'
fast/source/fast/buffer.d:125: error: function declaration without return type. (Note that constructors are always named 'this')
fast/source/fast/buffer.d:126: error: no identifier for declarator const length()
fast/source/fast/buffer.d:131: error: user defined attributes must be first
fast/source/fast/buffer.d:131: error: no identifier for declarator property
fast/source/fast/buffer.d:131: error: semicolon expected, not 'void'
fast/source/fast/buffer.d:143: error: user defined attributes must be first
fast/source/fast/buffer.d:143: error: Declaration expected, not '('
fast/source/fast/buffer.d:147: error: unrecognized declaration
```

gdc (Ubuntu 4.8.2-19ubuntu1) 4.8.2

kostya:
But even with dmd i got 0.43s vs RapidJSON 0.73s

kostya:
wow, how this is so fast?

mleise:
When I started I just scanned for object/array start and ends with SSE and measured the time. It was somewhere between 0.1s and 0.2s. Then when adding actual number parsing, string parsing and UTF-8 validation I kept a close eye on my time budget. Some of the performance is gained by doing heuristics first. While parsing numbers you can calculate exactly if an integer will overflow, or you can go faster by just comparing against a fixed value and still cover 99.99% of cases while handling the rest outside of the hot loop. Or you can add heuristics to the white-space checking function like RapidJSON did. SSE is fast when you have more than two bytes to skip, but white-space in JSON is often at most one character, so you better check that before doing SIMD processing. Things like force-inlining the white-space skipping function also made a 10% difference in overall speed at this level.
What I added was UTF-8 validation of strings coming from files, as I believe you should always do that on external input. But by far the biggest difference (factor of 2) is the parsing style. I know four ways to do it:

1. Build a DOM and work on JSON data by adding and removing nodes. Fields have no static type but some kind of "Algebraic" or "Variant". You can set a string field to 0.1 and it will change to a number field behind the curtain.
2. Pull parsing, where you can peek at the next value and then skip or consume it in one function call. For objects and arrays you pass a callback and continue processing in there. Fundamentally there is no document any more and processing is linear. On the other hand you get rid of dynamic types in statically typed languages like C++ or D and have the freedom to skip values you are not interested in. In particular it is faster to just validate a JSON number is correctly formatted than to convert it to a double.
3. Pull parsing, where you receive tokens in the form (type, text). An object would start with (Type.object, "{"). The structure is flattened and you need to keep track of the nesting levels. You can work with the raw strings though, which is great if you want to reproduce the original input.
4. Push parsing. Much like 3. but instead of you pulling things from the parser, you implement a set of functions for each type of token and the parser calls you back for each.

RapidJSON supports 1 and 4 as far as I can tell. 4 is way to cumbersome to use routinely and 1 adds unnecessary overhead in many cases. I decided to go with 2 and add a layer of convenience on top. For example I used a D's dispatch operator to make a JSON key lookup look like a property: `json.coordinates` gets rewritten into `json.singleKey!("coordinates")()`, a function that takes one compile-time argument and no run-time arguments and processes only that one key in an object while skipping over others. Only inside `read` I make use of D's garbage collector while building the dynamic array of `Coord` structures. The low memory consumption also stems from the fact, that I don't build a complete DOM and map the file into memory at once, often just reusing the disk cache as is.

On the downside I did not validate the unused side-structures. I think it is not necessary to validate data you are not using. So basically I only scan them so much as to find where they end. Granted it is a bit of optimization for a benchmark, but is actually handy in real-life as well. After all you could still raise the validation level to maximum if you really cared or call one of the validation functions.

Thanks for adding fast.json. Being the #1 feels good ;)

kostya:
great, you should create article from this and post somewhere.

miloyip:
I have fixes the read file part and provide SAX version for RapidJSON in #53 .

Since I do not have GNU D compiler on this machine yet and cannot do a comparison directly.
The new SAX version should be sure to spent much less memory, although writing such handler is a little bit clumsy. 

This comparison may be not fair in somehow, for example, RapidJSON test does not turn on UTF-8 validation, and the fast.json does not validate the parts that are not used. Should the tests be adjusted?

mleise:
Good to see you @miloyip. I hoped it would take a while before you notice that with SAX parsing you could do a lot better in this benchmark. Something is strange with the results though. I did try your push parser in a quick hack, just to see what to expect when you come around to improve your benchmark entry and it was more than twice as fast as the DOM parser ... on my i5.

As for the validation, it seems that everyone has a slightly different idea of what needs to be validated and as the authors of the parsers we know what kind of validation costs us precious milliseconds. My stance was that external input needs to be validated down to the character encoding level. But since my parser is effectively just a one-way deserializer, errors in unused parts don't matter, as they can't propagate.

We could certainly arrange for our two parsers to do the same level of validation, but what about the other entries? Some "read file as string" functions perform UTF-8 validation already, others don't. I'm actually in favor of keeping things as is, since not validating unused parts gives fast.json such a huge advantage, while still catching any errors that could invalidate the program output.

PR #49, opened 2015-10-22
add perl benchmarks for base64 and json, pure-perl and XS

Grinnz:
XS implementations are in C but they are normally what is used. Pure-perl benchmarks included for completeness. Requires at least perl 5.8.1. `wget -qO -` in build scripts can be replaced with `curl -L` if desired.


kostya:
thanks, perl xs quite fast for dynamic language.

PR #51, opened 2015-10-22
add perl benchmarks for matmul

rurban:
same speed as ruby


kostya:
thanks, the time is quite evil

PR #52, opened 2015-10-22
added Clojure JSON test

yogthos:
One thing to note is that due to the JVM warmup the initial run performance will not be representative of overall performance for a long running process. Normally, you'd do multiple runs with a warmed up JVM using something like [Criterium](https://github.com/hugoduncan/criterium).


kostya:
thanks, better to wget all this files in build.sh

yogthos:
The proper way to do it would be to use [leiningen](http://leiningen.org/) build tool, if you can set that up, I can package a proper project the way it's done for haskell and rust.

kostya:
would be nice

yogthos:
ok cool just updated the pr to use leiningen

PR #53, opened 2015-10-22
Improve RapidJSON performance

miloyip:
By using `FileReadStream` instead of C++ iostream converisons, less time was spent in reading the file. Also, provided a SAX version which is faster and cost much less memory.

On my rMBP:
~~~
Original
1.64s, 1031.5Mb

Using FileReadStream
0.75s, 336.3Mb

SAX
0.43s, 0.7Mb
~~~

### DIFF
--- a/json/build.sh
+++ b/json/build.sh
@@ -25,6 +25,7 @@ if [ ! -d rapidjson ]; then
   git clone --depth 1 https://github.com/miloyip/rapidjson.git
 fi
 g++ -O3 test_rapid.cpp -o json_rapid_cpp -Irapidjson/include
+g++ -O3 test_rapid_sax.cpp -o json_rapid_sax_cpp -Irapidjson/include
 
 if [ ! -d gason ]; then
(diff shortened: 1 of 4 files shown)


kostya:
after update rapidjson repo, 0.79 -> 0.94 (i not updated result for half of year), and sax result is 0.72

miloyip:
I think there should not have commits that affect the performance so much. The results are quite contradictory to the one on my machine. I am curious why an i7 machine performs worse than i5 machine though.

miloyip:
I suspect that maybe SSE was enabled previosly but FileReadStream disables it. And on your box that became a bottleneck. I will try it later.

kostya:
may be different gcc versions?

kostya:
problem in commit https://github.com/miloyip/benchmarks/commit/34f96c2f25556024f9095ddd1536654a23ac632f, if i revert it 0.92 -> 0.72

kostya:
read_file function executed in 0.12 for me.

PR #57, opened 2015-11-02
added Python3 ujson 

lqdc:
Because regular Python3 is a lot faster on this test than regular Python2.  And since it is not fair to only use python2 with ujson.


lqdc:
Note that Numpy is not some weird library. It is the main one used for matrix multiplication in Python, and nobody does it by hand in the C-way. So it makes sense to  use a library here if we are using libraries elsewhere (e.g. JSON). Even on http://julialang.org/ they use numpy for python matrix benchmarks.

PR #58, opened 2015-11-03
Removed unneeded file read to string for Nim JSON test

goldenreign:
This allows to save memory and decrease run time too.
Also please note that there is new Nim 0.12.0 release available: http://nim-lang.org/news.html#Z2015-10-27-version-0-12-0-released

### DIFF
--- a/json/test.nim
+++ b/json/test.nim
@@ -1,7 +1,6 @@
 import json
 
-let text = readFile("./1.json")
-let jobj = parseJson(text)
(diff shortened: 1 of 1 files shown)


kostya:
after update to nim 0.12, havlak example just crashed:
```
Welcome to LoopTesterApp, Nim edition
Constructing Simple CFG...
Error: unhandled exception: key not found: 0 [KeyError]
0.01s, 0.0Mb

```

PR #60, opened 2015-11-13
Array map instead of hash table

9il:
2-3 times faster


kostya:
all languages using hash map, so not fair to change it like this. but i going to rewrite it.

wilzbach:
@9il I just had the same idea (#87) :)

> so not fair to change it like this. but i going to rewrite it.

@kostya are you still planning to do this?
It would be fair to create a separate test for hash maps then...

kostya:

PR #63, opened 2015-11-16
Havlak Nim fix for 0.12.0 compiler version

goldenreign:
It fixes crash on Nim 0.12.0.
tables.[] behavior is the first in the list of breaking changes in http://nim-lang.org/news.html#Z2015-10-27-version-0-12-0-released
I changed only one of four occurences as other cases seem to behave correctly (exception if not found).


dom96:
:+1:

PR #65, opened 2015-11-26
Idiomatic

snadrus:
Float32 because it's all that NodeJS offers, so it must be good enough.
FFJSON would greatly improve this, if we're allowed to use things outside the standard library & generated code.
I can also rewrite with parallelism if interested, since that is one of the key reasons to use GoLang


kostya:
return float64, other implementations use it.

PR #67, opened 2015-11-26
Simpler types & Trivial parallelism

snadrus:


kostya:
please not change the algorithm, because need to implement the same in other languages. btw this is also not compiled

snadrus:
So you're only comparing where accessors are compiled away and where
parallelism is automatic?

The purpose of a language offering easy parallelism is that it's a choice
other than compiler optimizations that doesn't always require a difficult
rewrite.

please not change the algorithm, because need to implement the same in
other languages. btw this is also not compiled

—
Reply to this email directly or view it on GitHub
<https://github.com/kostya/benchmarks/pull/67#issuecomment-159921966>.

jadbox:
>> So you're only comparing where accessors are compiled away and where
parallelism is automatic?

I think keeping code fairly naive is a pretty good stance to take for these benchmarks, as doubling the size of the code for explicit parallelism seems to go too far imho.

PR #68, opened 2015-11-26
Read what we need, fold Tape into Run

snadrus:


kostya:
please return Tape struct, all implementations use it.

PR #74, opened 2015-12-29
Updated Go base64 to use concurrency

timstudd:
* Execution time reduced by ~50%


kostya:
thanks, but benchmark measured single cpu base64 convert. so not fair to compare it with other implementations.

PR #75, opened 2016-01-08
Improve the performance of the BF interpreter in Tcl

dbohdan:
The reworked interpreter is fast enough to run the `mandel.b` benchmark in reasonable time. This PR adds it to `brainfuck/run2.sh`.

### DIFF
--- a/brainfuck/brainfuck.tcl
+++ b/brainfuck/brainfuck.tcl
@@ -1,99 +1,111 @@
 package require Tcl 8.6
 
-namespace eval ::brainfuck {}
+namespace eval ::brainfuck {
+    variable verson 0.2.0
+    variable debug 0
 
-::oo::class create ::brainfuck::Interpreter {
-    constructor {} {
-        my variable tape
-        my variable pos
-
-        set tape [lrepeat 30000 0]
-        set pos 0
-    }
-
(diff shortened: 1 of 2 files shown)


kostya:
Thanks, but seems you changed the algorithm, which not correct to compare with other implementations.

Also i cant run this: `unknown or ambiguous subcommand "cat": must be bytelength, compare, equal, first, index, is, last, length, map, match, range, repeat, replace, reverse, tolower, totitle, toupper, trim, trimleft, trimright, wordend, or wordstart
    while executing`

dbohdan:
Regarding `string cat`: Your Tcl version is probably older than 8.6.2, which introduced it. It's easy to replace with a more widely supported command.

I understand why you would want to have the same algorithm implemented in different languages but I also think it limits the usefulness of the benchmark for languages that are substantially different from mainstream curly braces languages. For instance, in Tcl you can imitate the C++ or the Java way of implementing an interpreter. My previous implementation did that. However, the algorithm my new version uses is both faster and more concise because it uses the language more idiomatically. In Erlang or Caramel try to translate the C++ algorithm directly would likely be even worse for the code quality and the performance. 

I think a nice solution to this would be to have two versions of each benchmark: one with a fixed algorithm and another where everyone tries to make their implementation as far as the language allows.

PR #84, opened 2016-03-20
Use Crystal's SimpleHash in brainfuck

splattael:
It's in stdlib so always available.

### DIFF
--- a/brainfuck/brainfuck.cr
+++ b/brainfuck/brainfuck.cr
@@ -1,3 +1,5 @@
+require "simple_hash"
+
 class Tape
   def initialize
     @tape = [0]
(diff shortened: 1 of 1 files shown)


asterite:
@splattael Please no. See what I just wrote [here](https://github.com/crystal-lang/crystal/issues/2331).

I also benchmarked it, it's slower, so I don't think it makes much sense to change it :-)

splattael:
@asterite Thanks for your feedback :-)

My benchmark on brainfuck with `bench.b` showed a performance boost (6s -> 4.8s) after I did some profiling (with Linux `perf` tool).

As I am a newbie in crystal I fully trust you there :-)

PR #89, opened 2016-06-17
Many improvements to bf.d

JackStouffer:
* Use official D style as described here: https://dlang.org/dstyle.html
* Removed unnecessary calls to `std.conv.to`. In D, `char` implicitly converts to integer types, so no need to cast or convert `0` to `char`. Also, because brainfuck only contains ASCII values, we can safely cast from `int` to `char` with no loss of precision
* Removed semi-colons at the end of classes. These are not needed in D and just result in empty declarations
* Added attributes that allow the compiler to use better code-gen

### DIFF
--- a/brainfuck2/bf.d
+++ b/brainfuck2/bf.d
@@ -6,88 +6,164 @@ import std.conv;
 
 final:
 
-enum OpT { INC, MOVE, PRINT, LOOP };
-
-struct Op {
-  OpT op;
-  int v;
-  Op[] loop;
-
-  this(OpT t, int _v) { op = t; v = _v; }
-  this(OpT t, Op[] _l) { op = t; loop = _l; }
+enum OpT
+{
+    INC,
+    MOVE,
+    PRINT,
+    LOOP
 };
 
-class StringIterator {
-  string text;
-  int pos;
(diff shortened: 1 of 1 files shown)


kostya:
thanks, somehow after this d from 3th place move to 6th place

JackStouffer:
Odd. Are you using the same compilers with the same arguments?

kostya:
yes

kostya:
and with just switching branches i get difference, so it reproducable:
`old: 184d954e454309ba3edc5b1940862523226a1524, 2.37s, 1.4Mb`
`new: 3ef942d8c71bd0addec5624e4676c13cee441e8e, 3.07s, 1.4Mb`

JackStouffer:
How bizarre. What about the other compilers?

kostya:
https://github.com/kostya/benchmarks/commit/eccc0e54abec04547275b14acc0831996d00d4a8

kostya:
actually only attrs `@safe pure nothrow:` make this change, so i think i revert this commit ca4924432164f8123b144862046410865eb3b093

JackStouffer:
> make this change, so i think i revert this commit ca49244

That would be great :)

PR #90, opened 2016-06-22
Improve D base64 code

JackStouffer:
* Use official D style
* Make constants compile time constants
* Removed unnecessary GC use
* Removed unused variable
* Removed unnecessary conversion from msec to seconds

### DIFF
--- a/base64/test.d
+++ b/base64/test.d
@@ -19,7 +19,7 @@ int main()
         s += str2.length;
     }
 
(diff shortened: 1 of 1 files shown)


kostya:
thanks, but this https://github.com/kostya/benchmarks/pull/90/commits/aad883f0d8f65f3533814b644d33a4efe66027e9, output made bad, so i revert it.

PR #92, opened 2016-06-29
Add kotlin codes

mizunashi-mana:
Kotlin lang: https://kotlinlang.org/

Add:
 * brainfuck2
 * base64
 * matmul

### DIFF
--- a/brainfuck2/bf2.kt
+++ b/brainfuck2/bf2.kt
@@ -9,41 +9,25 @@ enum class OpT {
     INC, MOVE, PRINT, LOOP
 }
 
-class Op {
-    var op:OpT
-    var v:Int = 0
-    var loop:List<Op>
-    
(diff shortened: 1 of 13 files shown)


brainfuck2/bf2.kt
```diff
@@ -0,0 +1,116 @@
...
+        v = _v
+        loop = ArrayList<Op>()
+    }
+    
+    constructor(_op: OpT, _l: List<Op>) {
+        op = _op
+        loop = _l
+        v = 0
+    }
+}
```
kostya: thanks, maybe kotlin have some case class like in scala? to replace this.
https://github.com/kostya/benchmarks/blob/master/brainfuck2/bf.scala#L1

mizunashi-mana:
@kostya I fixed

PR #94, opened 2016-07-27
base64: build.sh: build C aklomp with OpenMP support

aklomp:
Build the base64-C-aklomp library with OpenMP support. The library has been significantly improved (made faster) since the first benchmark round. One of the improvements is the addition of OpenMP parallellization; this commit adds support for it. Should work out of the box on modern GCC and LLVM.

Disclaimer: I am the library's author.


kostya:
thanks, but i think this is not correct to compare with others because it multithreaded

aklomp:
I don't think multithreading is an unfair advantage for C, because other languages have the option to parallellize their implementation too. (They could even use my library :-) I think it's fair game. But I understand if you want to keep your benchmarks uniform.

Thanks for updating the non-multithreaded benchmarkl; good to see that the improvements paid off!

PR #96, opened 2016-08-04
Small header fixes

cwaffles:
- Fixed spacing for time
- Corrected megabit to megabyte (base 2)

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,199 +8,199 @@ Some benchmarks of different languages
 
 ### bench.b
 
-| Language        | Time,s  | Memory, Mb |
-| --------------- | ------- | ---------- |
-| D Ldc           | 2.02    | 0.9        |
-| C++ Gcc         | 2.25    | 1.0        |
-| Rust            | 2.53    | 4.9        |
-| Nim Gcc         | 2.56    | 0.7        |
-| Crystal         | 2.82    | 1.3        |
-| D Gdc           | 3.05    | 1.4        |
-| Nim Clang       | 3.09    | 0.8        |
-| Scala           | 3.43    | 120.12     |
-| Java            | 4.03    | 513.8      |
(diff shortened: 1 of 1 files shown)


kostya:
hm,

```
MiB = Mebibyte = 1024 KiB
KiB = Kibibyte = 1024 Bytes
MB = Megabyte = 1,000 KB
KB = Kilobyte = 1,000 Bytes
```

and this should be MB, because megabytes

cwaffles:
Linux counts memory usage in base 2, meaning a KiloByte is 1024 bytes.

See this post for example http://www.tecmint.com/check-memory-usage-in-linux/.

The amount of RAM used is 1046147072 bytes. If you want to convert that to Megabytes (base 2), divide by 1024^2. Which is 997.68, which matches the conversions later on.

However if you divide it by SI bytes (base 10), then the result is 1046, which doesn't match the conversions later on.

Networking is base 10, hard drive storage is mixed between base 2 and base 10, and RAM is base 2.

From wikipedia:

> The hardware industry has coped with the dual definitions because of relative consistency: system memory (RAM) typically uses the binary meaning while magnetic disk storage uses the SI meaning. There are, however, exceptions and special cases. Diskettes use yet another "megabyte" equal to 1024×1000 bytes.[37] In optical disks, Compact Disks use MB to mean 10242 bytes while DVDs use GB to mean 10003 bytes.[38][39] -- https://en.wikipedia.org/wiki/Binary_prefix#Operating_systems

kostya:
i measuring like this:
```
ps p 63009 -o rss
  RSS
141736

141736 / 1024.0
=> 138.41 Mb
```

cwaffles:
Yes, which is a MegaByte in base 2. I wrote MiB to avoid confusion between base 2 and base 10 systems.

kostya:
not understand really, but ok

PR #105, opened 2016-09-22
Mir update: cache block sizes was improved for MatMul

9il:
also makeUninitializedSlice was added


9il:
It should be as fast as single thread Julia +/- 0.01 sec.

9il:
Thanks @kostya, just finished my benchmark: [Mir GLAS vs OpenBLAS](http://blog.mir.dlang.io/glas/benchmark/openblas/2016/09/23/glas-gemm-benchmark.html). It may be interesting for you.

PR #110, opened 2016-11-06
Kotlin version is rewritten like in Scala using case classes and arrays

elizarov:


kostya:
wow how this can be faster than c?

elizarov:
Of course, you can also rewrite both C and Java code is similar vein to get similar performance. It will become more complicated, though, beucase Scala and Kotlin support sealed/cases classes natively in their object models, so this is a natural code style for them, but unnatural one for C and Java.

PR #113, opened 2016-12-20
Added functional programming version of go

lologarithm:
Made a version of the go interpreter that matches the Scala implementation instead of the standard struct based system.

Performance is better as well :)

### DIFF
--- a/brainfuck2/bf_func.go
+++ b/brainfuck2/bf_func.go
@@ -1,0 +1,131 @@
+package main
+
+import "fmt"
+import "os"
+import "io/ioutil"
+
+const (
+	INC = iota
+	MOVE
+	PRINT
+	LOOP
+)
+
+type Op func(*Tape)
+
+type StringIterator struct {
+	Text []byte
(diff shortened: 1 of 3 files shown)


kostya:
hm seems this is slow for bench.b, 5.36s -> 5.57s, for mandel.b it faster 100s -> 70s, but for gccgo it also slower.

lologarithm:
That seems strange it would be slower for bench.b but faster for mandel.b. 

I didn't test with gccgo locally (only go 1.7.3 on ubuntu 16.04).

I can do some more benching and see if I can figure out why that would be. Could you try a few times and see if that time was just a fluke?

lologarithm:
mandel.b
Go
75.26s, 2.7Mb
Functional Go
48.05s, 1.7Mb

For bench.b I see
Go
4.34s, 1.4Mb
Functional Go
3.10s, 1.4Mb


I ran the tests about 10 times and didn't see much divergence from what I pasted here.

PR #122, opened 2017-02-15
Repeat benchmarks to eliminate noise

### LINKED ISSUE #121: Repeat benchmarks to eliminate noise
I can run the same benchmark a few times and get wildly different results. Consider having xtime.rb loop 10-100 times and take the minimum to filter out some of this noise.

dtolnay:
Fixes #121.


kostya:
it would be much slower to run, so i think no edit xtime.rb, but create new check.rb.

dtolnay:
Makes sense. I updated the PR with repeat.rb and kept the original xtime.rb.

PR #124, opened 2017-02-15
Memory map the JSON file in Rust

dtolnay:
This is an 18% improvement on my computer. The D code does the same thing so it's only fair!

cc @pmarcelll


kostya:
d still faster :)

pmarcelll:
That is probably due to the use of explicit SIMD.

PR #131, opened 2017-05-14
Matmul with less indexing and more iterators

leonardo-m:


leonardo-m:
On my PC I've found `-C opt-level=3 -C lto` compiler switches to be better for this program.

PR #134, opened 2017-06-14
Matmul with D-style dot_product

leonardo-m:
A new matmul that uses a dot_product function that's inspired by the D
language library function.

For me the best compiler switches for this program are:
-C opt-level=3 -C target-cpu=native


leonardo-m:
This is how the D standard library does it:
https://github.com/dlang/phobos/blob/master/std/numeric.d#L1715

Do you suggest me to close down this issue?

PR #140, opened 2017-08-07
Use triple-loop for Julia matmul implementation

simonbyrne:
The earlier Julia code allocated a lot of temporaries: this uses a fairly simple triple loop (similar to C and other languages).


kostya:
in my run, this is slower 27s vs 20s

simonbyrne:
That's bizarre. I got 12 seconds vs 44 seconds. Anyway, I've updated it now to do the pre-transpose, similar to C, as well as some annotations to disable bounds checking, and I get 4 seconds.

kostya:
super.

PR #146, opened 2017-10-05
[WIP] add json asdf

9il:
LDC ~ 1 sec
DMD ~ 1.5 sec

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,11 @@
 *~
 .dub/
 target/
+*.o
+*.s
+*.a
+.DS_Store
+json/test_asdf
+json/test_mir_asdf
+json/test_mir_ion
+json/test_mir_ion_file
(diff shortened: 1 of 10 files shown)


json/test_asdf.d
```diff
@@ -0,0 +1,54 @@
+/+dub.json:
+{
+    "name": "json_d_asdf",
+    "dependencies": {
+        "asdf": {"path" : "../../asdf"}
```
tchaloupka: Should be package name and version from the dub repository instead

kostya:
not builded
```
dub build --compiler=dmd --build=release-nobounds --single test_asdf.d && mv json_d_asdf json_d_asdf_dmd
No package file found in /home/kostya/projects/asdf/, expected one of dub.json/dub.sdl/package.json

```

json/test_asdf.d
9il: Ah, yes. OK, i will fix it after the next asdf beta. plz wait with this PR

PR #149, opened 2017-12-19
bf.lua: localised more variables

liuchia:
Rewrote bf.lua to reduce amount of table indexing. Now in similar style to the javascript benchmark.

Made data and pos local as they should not be public. Localized loop_code in parse (No noticeable performance benefit but good lua practice).

Performance on my machine: 
```
BENCH  luajit original : 16.26s 1.1Mb
BENCH  luajit new      : 11.27s 1.1Mb

BENCH  lua5.1 original : 170.68s 1.1Mb
BENCH  lua5.1 new      : 123.36s 1.1Mb

MANDEL luajit original : 133.88s 2.0Mb
MANDEL luajit new      : 116.37s 2.0Mb
```

### DIFF
--- a/brainfuck2/bf.lua
+++ b/brainfuck2/bf.lua
@@ -3,75 +3,88 @@ local MOVE = 1
 local PRINT = 3
 local LOOP = 4
 
--- Tape class
+local function Tape()
+  local pos = 1;
+  local data = {0};
 
-local Tape = {}
-Tape.__index = Tape
-
-function Tape.new()
-  local self = setmetatable({}, Tape)
-  self.data = {0}
-  self.pos = 1
-  return self
-end
-
-function Tape.current(self)
-  return self.data[self.pos]
-end
-
-function Tape.inc(self, x)
-  self.data[self.pos] = self.data[self.pos] + x
-end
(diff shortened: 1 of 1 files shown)


TSnake41:
You can get an even faster code by optimize table.insert by table definitions.
`lua
local function Brainfuck(text)
  local function parse(source, i)
    local res, count = {}, 1;
    while i <= source:len() do
      local c = source:sub(i, i);
      if c == "+" then
        res[count] = {INC, 1};
        count = count + 1
      elseif c == "-" then
        res[count] = {INC, -1};
        count = count + 1
      elseif c == ">" then
        res[count] = {MOVE, 1};
        count = count + 1
      elseif c == "<" then
        res[count] = {MOVE, -1};
        count = count + 1
      elseif c == "." then
        res[count] = {PRINT};
        count = count + 1
      elseif c == "[" then
        local loop_code;
        loop_code, i = parse(source, i+1);
        res[count] = {LOOP, loop_code};
        count = count + 1
      elseif c == "]" then
        break;
      end
      i = i + 1;
    end
    return res, i;
  end`

PR #153, opened 2018-01-18
Change brainfuck.nim to check if char is in set and not string

hlaaftana:


kostya:
thanks, but this is actually old benchmark, on main page used brainfuck2 benchmark, and also improve parsing useless, because it use 1% of time.

PR #156, opened 2018-06-11
Rust: Remove useless stdout flushing

Kerollmops:
Flushing is not mandatory here and I suppose Crystal do buffering to gain performances, so it is more fair to let Rust do automatic flushing to avoid too many system calls too (`stdout` flushing explicitly call `write` from the `libc`).


kostya:
all languages use flush, this is required, and also there small amount of puts actually, so not related to perfomance at all.

PR #157, opened 2018-07-10
Go: Remove unnecessary "t.move(1)" in NewTape; add "default" case in parse

wpwoodjr:
Without a "default" case, parse was adding an extra op (inc 0) to the program whenever the input had a non-matching character


nuald:
Applied in the another commit.

PR #160, opened 2018-07-13
Performance improvments for brainfuck Go program

devlo:
Changed range loop to normal for, eliminated a lot of copying.
	
Before:	User time (seconds): 4.50
After: 	User time (seconds): 2.22

Twice faster now.

### DIFF
--- a/brainfuck2/bf.go
+++ b/brainfuck2/bf.go
@@ -103,14 +103,20 @@ func parse(si *StringIterator) []Op {
 }
 
 func _run(program []Op, tape *Tape) {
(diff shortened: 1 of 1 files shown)


cristaloleg:
By the way, this case is handled by [go-critic](https://github.com/go-critic/go-critic) in `rangeValCopy` checker (and [docs](https://go-critic.github.io/overview#rangeValCopy-ref)).

kostya:
sorry, right now not have access to test machine, btw, i expected this loop should be optimized by itself.

PR #161, opened 2018-09-22
Increase Go performance

kempjeng:
* Loop via index, it avoids a copy

* Append all at once rather than a loop
It uses the same allocation strategy that is used in the F# implementation.

* Don't use `printf` it runs through a parser.
I think this PR is slightly more obtuse.  Perhaps a `print(byte(tape.Get()))` would be clearer.

Original time:  119s
After modifications: 21s

5.6x increase in performance.


kostya:
merged in https://github.com/kostya/benchmarks/pull/160

PR #167, opened 2019-01-23
Update Rust base64

pickfire:
Before and after (10 tries)

```
encode: 133333360, 9.954792112
decode: 100000000, 10.080134839
```

```
encode: 133333360, 1.776409588
decode: 100000000, 8.38234092
```


nuald:
Applied in the another commit.

PR #172, opened 2019-04-30
python base64 improvement

ChristianHarms:
Like perl I would add a base64 module for encode/decode function to get better performance for python. Factor10 for python is available.


ChristianHarms:
Before:
encode: 1333333600, 2.20062494278
decode: 1000000000, 3.34313797951

After:
encode: 1333333600, 0.165750980377
decode: 1000000000, 0.427265882492

kostya:
LGTM

PR #173, opened 2019-06-29
Better Bf2 Java implementation.

renatoathaydes:
This PR implements brainfuck2 in Java that runs much faster than the current implementation.

The code is still pretty clean and very readable, besides being [more testable](https://github.com/renatoathaydes/bf-jvm/blob/master/src/test/kotlin/BfTest.kt).

On my system, here are the figures I am getting comparing the new implementation with the current one, as well as with the Kotlin (current fastest in the project README) implementation:

> * in seconds

| Java (current) | Java (new) | Kotlin |
|------------------|-----------------|--------|
| 2.6 | 1.4 | 2.0 |

The Kotlin implementation is also not optimal, I will make a PR offering a faster implementation for it as well.

### DIFF
--- a/brainfuck2/bf.java
+++ b/brainfuck2/bf.java
@@ -1,135 +1,149 @@
 import java.io.IOException;
+import java.io.PrintStream;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
-import java.util.Stack;
+import java.util.PrimitiveIterator;
+import java.util.function.Consumer;
 
-public class bf {
 
-    public enum OpT {INC, MOVE, PRINT, LOOP}
+final class bf {
 
-    public static class Op {
-        public OpT op;
-        public int v;
-        public List<Op> loop;
-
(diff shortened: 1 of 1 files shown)


renatoathaydes:
I need to clarify that the times I posted above is when running Java 8 with the GraalVM's JIT.

On the Oracle JVM, there's a performance issue with the Java version, so it runs much slower than Kotlin.

Here's my data:

### Java HotSpot(TM) 64-Bit Server VM (build 25.151-b12, mixed mode)

- Java (new): first run and subsequent runs: 3.6 seconds
- Java (current): first run: 4.3 seconds. Subsequent runs: 3.0 seconds
- Kotlin: first run: 1.2 seconds. Subsequent runs: 1.55 seconds.

### OpenJDK 64-Bit GraalVM CE 19.0.2 (build 25.212-b03-jvmci-19-b04, mixed mode)

- Java (new): first run: 1.48 seconds. Subsequent runs: 1.43 seconds
- Java (current): first run and subsequent runs: 2.5 seconds.
- Kotlin: first run: 2.0 seconds. Subsequent runs: 1.96 seconds.

So, the performance depends a lot on which JIT is used for Java.

renatoathaydes:
For comparison, the Go implementation on my machine runs in about 2.0 seconds.

PR #175, opened 2019-07-26
Greatly improve Java bf2 performance

sebthom:
By doing some very simple changes to the existing Java implementation (using array instead of list, using final modifierers, avoiding Character/char autoboxing, reusing empty arrays) I can see a 100% performance improvement on my system using OpenJDK 8 and OpenJDK 11.
This version apparently also runs significantly faster than version offered via https://github.com/kostya/benchmarks/pull/173


nuald:
Applied in another commit.

PR #177, opened 2019-08-13
nim: matmul uses less memory

narimiran:
Updated the code to be compatible with the latest Nim version.

In Nim 0.20, `-d:danger` is what `-d:release` used to be: updated the build scripts accordingly.

-----

Results for `./xtime.rb ./matmul/matmul 1500` on my machine:
* old: `-143.50017; 3.85s, 120.9Mb`
* new: `-143.50017; 3.80s, 73.2Mb`

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,7 +241,7 @@ Intel(R) Core(TM) i5-2400 CPU @ 3.10GHz (Ubuntu 14.04.1 LTS x86_64)
 # Versions:
 
 * gcc (Ubuntu 4.8.2-19ubuntu1) 4.8.2
-* Nim Compiler Version 0.17.0 (2017-05-17) [Linux: amd64]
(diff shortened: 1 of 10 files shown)


nuald:
Applied in another commit.

narimiran:
> Applied in another commit.

You didn't change `-d:release` to `-d:danger`.

nuald:
I have some doubts about the flag, but after checking the official documentation it seems that it makes sense as runtime checks add overheads (especially visible for tests operating a lot of array get/set operations) that other compilers don't have. Addressed in #182.

narimiran:
> I have some doubts about the flag

I can see the suspicion, but basically:
If the old tests with Nim v0.19 and below used `-d:release` (which they did), that directly translates to `-d:danger` in newer Nim versions (v0.20 and beyond).

Issue #179, opened 2019-09-16
Sad this isn't being updated anymore...


lbmn:
This was one of the best programming language benchmarks I was able to find anywhere!  (As ye olde [Benchmarks Game](https://benchmarksgame-team.pages.debian.net/benchmarksgame/) has made itself irrelevant by excluding top contenders like D and Nim.)  But performance competition is very dynamic, and a benchmark can only be relevant if it reflects the latest versions.  Very sad that the results haven't been updated in 200 days...

nuald:
I'm going to help with the project in my spare time, closing this ticket as less relevant now.

Issue #170, opened 2019-02-22
Please upgrade LDC / DMD / GDC version to latest


zoujiaqing:
LDC latest version is 1.18.0
DMD latest version is 2.088.0
GDC latest version is 9.2.0

Thanks for your work!!!

lbmn:
DMD is now at 2.08**_5_**.

Please also upgrade Nim as well as its gcc / clang backends to latest stable versions.

The race between Nim and Dlang is most interesting!

It would also be interesting to try Nim with more backends, like [icc](https://en.wikipedia.org/wiki/Intel_C%2B%2B_Compiler), [pcc](https://en.wikipedia.org/wiki/Portable_C_Compiler), [tcc](https://en.wikipedia.org/wiki/Tiny_C_Compiler), etc.

Issue #164, opened 2018-12-14
Update Kotlin


napperley:
With the benchmark a very out of date version of Kotlin is used (**1.0.3**). Please update Kotlin to the latest stable version (**1.3.11**).

Issue #159, opened 2018-07-13
truffle ruby with latest version. Graal 1 RC3


claudiug:

claudiug:
now already RC8

claudiug:
RC10

PR #180, opened 2019-10-10
The full suite update

nuald:
I know that you have problems with accessing to the test platform you've used previously and/or don't have time for the updates, so I did the full update and re-tested everything on my home server. Please feel free to communicate with me regarding additional updates and tests if needed. The details regarding the update are below.

Updated the toolchain (addressed tickets #129, #133, #158, #159, #164, #170, #179).

Removed Ruby Topaz and Rubinous as these are not the active projects anymore (the latter has some activity, but it neither can be properly compiled nor compatible with the used Ruby code).

Fixed memory consumption calculation to include children processes (e.g. for Scala as it invokes the actual JVM process as a child).

Minor tweaks to BF tests (normalizing data types and allocations if possible).

Applied changes from PR #157, #167, #177, #175.

Added relevant warming up for .NET projects (issue #154), however it barely changed the results.

Replaced D Mir GLAS (not supported anymore) with D lubeck (another linear algebra library also utilizing D Mir) test.


kostya:
Wow, that huge work, thanks, i will look at this shortly. Maybe you have something like Dockerfile for  easy to install all this languages?

nuald:
I wish I could utilize Docker, but unfortunately virtualized environments make the benchmarks misleading (for example, SSE2 optimizations may not work properly). I have my own benchmark for web performance (https://github.com/nuald/simple-web-benchmark), and you could see there that WSL (semi-virtualized Linux in Windows) gives completely different results comparing with actual Linux (you could note that average response times in WSL are almost the same for all languages, but on the actual hardware the average values are different).

It would be nice to have a tool that uses something like Dockerfile to create an actual bootable Linux ISO, but I couldn't find any project alive for that. However, if you think that Docker-based benchmarks could give us some valuable metrics, I can take a look to it. Please let me know what do you think.

nuald:
Another option could be creating some installation script that would fetch and update the packages, but some languages used in the benchmark are not trivial to install (like V8's D8 - it requires building from sources with the special toolchain installed). I guess nobody cares for D8 as nodejs is mature and fast enough, but I didn't want to reduce the number of languages. However, it's still an option - the majority of the languages are installed by either download/extract or "apt install X" commands.

kostya:
In Windows you may be would have difference, but in linux docker should transparently run process on host OS just with some kind of hidden namespace. I think this would be ok for this benchmark.

nuald:
Good to know, I wasn't aware that Docker uses LXC (unfortunately, GNU/Linux is not my main OS so I have some knowledge gaps). I'll start working on the Dockerfile.

brainfuck2/bf.cpp
```diff
@@ -29,7 +29,7 @@ class Tape {
 
   inline int get() { return tape[pos]; }
   inline void inc(int x) { tape[pos] += x; }
-  inline void move(int x) { pos += x; while (pos >= tape.size()) tape.push_back(0); }
+  inline void move(int x) { pos += x; while (pos >= tape.size()) tape.resize(2 * tape.size()); }
```
kostya: looks like little hack, i prefer minimum hacks

kostya:
All looks ok, thanks, i added you to project, feel free to merge, and close issues.

nuald:
Sorry, I doesn't seem that I have the access rights to either merge the PR or close the issues.

kostya:
seems you need to accept it on https://github.com/kostya/benchmarks/invitations

brainfuck2/bf.cpp
kostya: yes, but still push_back is a public api, its readable and understable call (i'm crystal coder in mind :) ). It should be language care to optimize it as fast as possible, not user. In other implementations, it was users PRs where i also prefer for simplify things.

PR #182, opened 2019-10-11
Applied danger flag to nim tests

nuald:
The flag seems fair as release flag have the runtime checks not occurring in other languages.


kostya:
seems ok

kostya:
or not, java, crystal for example have all runtime checks (int overflows, array bounds)

nuald:
Unfortunately, we can't be sure whether the runtime checks are applied or not in all the cases - it seems that certain heuristics could be used by JIT and/or compilers to remove the runtime checks in the final code based on the algorithms used (I suspect Kotlin applies some optimizations regarding that otherwise it's quite difficult to explain its miracle performance for bench.b). The fair comparison would be to compare all the code being optimized (as anything not optimized would be up to the compiler authors - some of them assume that not optimized includes debugging info provided, and at the same time others assume that non-optimized means actually optimized, but no aggressive methods like inlining are applied). So personally, I tend to accept some optimization applied for the benchmarks (unless it's some cryptic compiler flags not usually used/known by majority of developers).

PR #184, opened 2019-10-18
Update with changes to Nim's base64 API.

treeform:
lineLen was removed, to make it go faster.

### DIFF
--- a/base64/test.nim
+++ b/base64/test.nim
@@ -9,9 +9,7 @@ var t = times.epochTime()
 var i = 0
 var s:int64 = 0
 while i < TRIES:
-  # set large line length to avoid adding line breaks,
-  # outputing identical encoded strings as in other languages
-  str2 = base64.encode(str, 2 * STR_SIZE)
+  str2 = base64.encode(str)
   s += len(str2)
   i += 1
 echo("encode: ", s, ", ", formatFloat(times.epochTime() - t, ffDefault, 6))


nuald:
It looks like a fair change, and works faster indeed. As we temporarily use my home server for the testing, I've already got the results, and you may want to update the README with numbers:

- Nim Gcc: 3.63s, 58.9Mb
- Nim Clang: 3.74s, 59.2Mb

(Please move the rows accordingly as the tables are sorted by Time value).

treeform:
Your numbers don't appear to be faster. Are all numbers made with your server? 

Did you `choosenim devel` and or `choosenim update devel` to get the most recent version of nim?

In my machine (windows 64bit) it puts it between `C aklomp SSSE3` and `C` at 1.66s

nuald:
It's twice faster than before (as the current tests shows 7+ seconds). As for the absolute numbers - we can't compare your and my numbers as we have different software and hardware. Nim version on Windows could have some optimizations, but I don't have the dual boot PC to verify OS differences.

I'd like to note though that the devel version (the most recent one - 1.0.99) is faster comparing with the stable one (1.0.0, its numbers in my previous comment):
- Nim Gcc: 2.31s, 58.9Mb
- Nim Clang: 2.62s, 59.3Mb

But we don't use devel versions unless absolutely necessary (as the performance improvement one day could introduce the regression bug another day and they could revert that improvement).

nuald:
I've verified the tests in Windows 10 and its WSL (with ArchLinux distribution). While WSL won't provide the same performance as the dedicated installation, the relative numbers could be good enough for the comparisons. As I suspected, Nim behaves differently in different OSes and it may explain your numbers.

Windows environment:
- gcc.exe (Rev2, Built by MSYS2 project) 9.2.0 
- clang version 9.0.0 (https://github.com/msys2/MINGW-packages.git fdafa4d8c4022588676c8ec0985dafaf834258ae)
- Nim Compiler Version 1.0.99 [Windows: amd64]

WSL environment:
- gcc (GCC) 9.2.0
- clang version 9.0.0 (tags/RELEASE_900/final)
- Nim Compiler Version 1.0.99 [Linux: amd64]

Testing notes:
- xtime.rb doesn't support Windows, so for the fair results I've used the time values from the output of the binaries;
- Nim Clang is compiled without .exe extension on Windows, therefore I had to rename it.

Language | Windows 10 | WSL (ArchLinux)
-- | ----------- | -------------
C aklomp SSSE3  | 1.58 (0.88 + 0.70) | 0.93 (0.45 + 0.48)
C | 2.68 (1.30 + 1.38) | 3.26 (1.34 + 1.92)
Nim Gcc | 4.29 (1.93 + 2.36) | 9.04 (3.79 + 5.25)
Nim Clang | 4.55 (1.93 + 2.62) | 10.13 (4.44 + 5.69)

Please note the that the relative results are the same (Nim being slower than C), but the absolute values could differ a lot. I'm still surprised that you got faster than C results though.

PR #186, opened 2019-10-19
Make Nim work with new and old base64 API.

treeform:
Old version of Nim also did MIME encoding, which needs to be turned off.
* Also used `block:` to prevent timing from polluting each other.
* Also used `for` loop instead of a `while` loop.

### DIFF
--- a/base64/test.nim
+++ b/base64/test.nim
@@ -5,19 +5,22 @@ let TRIES = 100
 let str = strutils.repeat('a', STR_SIZE)
 var str2 = ""
 
-var t = times.epochTime()
-var i = 0
-var s:int64 = 0
-while i < TRIES:
-  str2 = base64.encode(str)
-  s += len(str2)
-  i += 1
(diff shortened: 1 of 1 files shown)


nuald:
This change barely improves performance - as default `newLine` in `base64.encode(str)` is the empty string anyway, it doesn't affect the results that much (I've got encode time for previous code version 1.17s, and the new version - 1.15s for GCC; Clang results have not been changed at all).

As for this particular pull request, I'm fine with `block` and `for` changes, but would avoid any dependencies on the Nim version (especially considering that it doesn't change the test results). So if you submit the new change without that dependency, I'll merge it. As for the better performance in the devel version, you may influence Nim developers to get the new stable release rather sooner, and after that we could use it and update the results.

However, if @kostya decides that using Nim devel version is fine in the benchmark, I can update the results with it.

treeform:
You right the default is an empty string. I made a mistake there. I don't think my block or for loop are that valuable. Lets just close it.

PR #190, opened 2019-10-22
Improved build process

nuald:
As a part of preparation for Docker usage I had to improve the build process and remove the hidden dependencies, therefore:

- Removed SBT usage as can't use the external scalac compiler (the new build process is based on https://github.com/jodersky/scala-makefile).
- Removed Maven usage as it doesn't use the external javac compiler.
- Replaced Haskell Stack with Cabal as the former doesn't use the external ghc compiler.
- Removed Leiningen usage as it doesn't use the installed Clojure.

Also removed unused flag -opt:l:inline for Scala (it works only in if -opt-inline-from is used, and adding the latter actually decreased performance for some tests).


kostya:
about Docker, i saw nice example here: https://github.com/drujensen/fib/blob/master/Dockerfile

nuald:
Thanks for the link, it looks close to what I have (still work in progress, submitted just for the reference): https://github.com/nuald/benchmarks/blob/docker/Dockerfile

Few notes though:
- initially I also used Debian base, but both Ubuntu and Debian packages (including testing) are quite behind the current (even stable) releases of the languages. It's mostly critical for Gcc and Clang - it takes hours to compile them and Docker occasionally fails during that. I've switched to Arch Linux as its packages are usually updated more frequently.
- https://asdf-vm.com/ looks very interesting, I didn't know about it; however Arch has AUR packages that has the same meaning - community-driven installation scripts. Unfortunately, asdf doesn't contain the most problematic to install languages like Swift and .NET's (I guess because their authors don't bother so much about Linux support);
- using generic-purpose programming language instead of bash for builds and runs is tempting, but as the current scripts are already working, I don't see the reason to meddle with it on this stage. Surely, you can go ahead and implement it, I'll update the Dockerfile accordingly later. For my other benchmarks I also used the generic-purpose programming language (Scala) as bash isn't flexible enough, has quite cryptic syntax and rely on tools which may have the same problems with syntax (and moreover, could be not portable between Linux and BSD-based systems like macOS).

PR #193, opened 2019-10-24
Optimize memory allocation for base64 tests.

nuald:
Rewrote C++ version to operate BIO directly.
Changed aklomp usage to utilize AVX and AVX2 CPU extensions if available.


kostya:
looks nice, thanks

PR #195, opened 2019-10-29
Improve V base64 memory usage by using pre allocated buffers.

spytheman:
Before this PR, on my PC:
```shell
0[13:21:51] /v/benchmarks/base64 LINUX $ ###BEFORE
0[13:22:00] /v/benchmarks/base64 LINUX $ /v/nv/v -prod -cc gcc -o base64_v_gcc test.v
0[13:22:08] /v/benchmarks/base64 LINUX $ ../xtime.rb  ./base64_v_gcc
encode aaaa... to YWFh...:   1464 ms | 1431666688
decode YWFh... to aaaa...:   2278 ms | 1073741824
3.84s, 2397.9Mb
0[13:22:14] /v/benchmarks/base64 LINUX $
```

After this PR, on my PC:
```shell
0[13:22:25] /v/benchmarks/base64 LINUX $ ###AFTER
0[13:22:35] /v/benchmarks/base64 LINUX $ /v/nv/v -prod -cc gcc -o base64_v_gcc test.v
0[13:22:40] /v/benchmarks/base64 LINUX $ ../xtime.rb  ./base64_v_gcc
encode aaaa... to YWFh...:   1095 ms | 1431666688
decode YWFh... to aaaa...:   2009 ms | decode: 1073741824
  3104 ms | <=== total time spent running base64 encode/decode ops
 ok, fail, total =     2,     0,     2
3.11s, 2.0Mb
0[13:22:46] /v/benchmarks/base64 LINUX $
```

### DIFF
--- a/base64/test.v
+++ b/base64/test.v
@@ -7,28 +7,33 @@ fn main() {
 
   str := 'a'.repeat(str_size)
 
-  mut str2 := base64.encode(str)
+  str2 := base64.encode(str)
+  str3 := base64.decode(str2)
+  
+  str2_buffer := malloc( str2.len )
+  str3_buffer := malloc( str3.len )
+  
   print('encode ${str.substr(0, 4)}... to ${str2.substr(0, 4)}...: ')
 
   mut bench := benchmark.new_benchmark()
   bench.step()
   mut s := 0
   for i := 0; i < tries; i++ {
-    str2 = base64.encode(str)
-    s += str2.len
+    s += base64.encode_in_buffer(str, mut str2_buffer)
   }
   bench.ok()
   println(bench.step_message('$s'))
 
-  mut str3 := base64.decode(str2)
   print('decode ${str2.substr(0, 4)}... to ${str3.substr(0, 4)}...: ')
 
   bench.step()
   s = 0
-  for i := 0; i < tries; i++ {
-    str3 = base64.decode(str2)
-    s += str3.len
+  for i := 0; i < tries; i++ {    
+    s += base64.decode_in_buffer( str2, mut str3_buffer )
   }
   bench.ok()
-  println(bench.step_message('$s'))
+  println(bench.step_message('decode: $s'))
+  
+  bench.stop()
+  println(bench.total_message('running base64 encode/decode ops'))
 }


kostya:
i not sure this is correct, comparing to others, all implementations use heap allocations (except c which use stack allocation).

nuald:
@kostya is right, however it's obvious that the test has the problem with the memory allocation. I've looked into the generated C code and yes, the allocated buffer is not freed, therefore it has the performance problem (as `malloc` can't reuse the previously allocated memory with `sbrk` as we didn't free it). I've created a PR #196 to address it and it definitely shows the performance improvements.

kostya:
fixed by #196

spytheman:
As it is, the v test measure the speed of malloc/free more than it measures actual base64 encoding/decoding.

See commit #befeaf855d59e6ecde9fa3de264ccbd3d1b48342 , where one of the main optimizations for multiple languages was making the hot loops use a preallocated buffer in one way or another, and eliminating calling malloc/free inside the loops.

Calling manually `free` fixes a memory leak with the current language implementation, but calling base64.decode_in_buffer instead of base64.decode and base64.encode_in_buffer instead of base64.encode , does the same as befeaf855 .

spytheman:
Should I make another PR, that replaces calls to base64.encode with base64.encode_in_buffer ?

nuald:
@spytheman `malloc`/`free` on small sizes are virtually free as they use `sbrk` (so it should reuse the same memory, just move the pointer). If you measure the performance of the current version and the version using `base64.encode_in_buffer` you won't see any difference.

However the latter one would be unfair in the general use-case as all other tests allocates the memory in the loop (as we measure "encoding/decoding into a new buffer" as it's not possible to measure just "encoding/decoding" for high-level languages as they don't support pre-allocated buffers). However, if V would support allocating on stack, it would be nice as in that case it would be possible to have closer-to-C code.

spytheman:
@nuald , V does support allocating on the stack, just do: `buf := [1000]byte` and you will get stack allocated 10000 bytes. In this regard, v is closer to C and should have similar performance.

nuald:
Looks like the stack allocation doesn't work with the variable sizes, I'm getting compilation errors like:
```
test.v:30:34: undefined: `byte`
   28|            str2_size := 4 * ((str_size + 2) / 3)
   29|            str2 := [str2_size]byte
                                        ^
```
No errors if use numeric values.

Please note though that the memory manipulations barely change anything here - it's not the main culprit here. aklomp uses SSE instructions, pure C version uses some optimizations too like `__builtin_bswap32`, the V base64 module doesn't have any optimizations though. It's still the decent results, just please don't expect it to be on par with the optimized versions (yet?)

PR #198, opened 2019-10-30
Docker support

nuald:
Big change introducing Docker-based tests. Few notes:
- the image is quite big (14.7 GB for me). I wanted to use Arch Linux base as it has the smaller footprint and has the newer base packages like GCC, but certain applications (in particular, Mono) are not up-to-dated there and had to be built from sources (and it fails too often in Docker as apparently it has a problem with many programs running at once writing to file system);
- the build time is about an hour (on my server) - the list of dependencies is big and some languages have to be built from sources. Occasionally (especially building from sources) it fails, but trying to build again usually works;
- the language versions are now the table;
- certain tests have to be run manually (it's noted in the README.md);
- Q, Felix, Rock and V8 (D8) tests have been removed (Q can't be installed automatically, Felix and Rock are not maintained anymore and their results are not used anyway, V8 is way too slow to compile and on par with Node.js anyway);
- if the PR is merged, I'm going to use the Docker-based numbers henceforth (unless you want to run the tests by yourself).

My suggestions for the future work (mostly regarding automating the tests; sorted by LoE):
- reduce the JSON file size (as it may make the JRuby and TruffleRuby tests stable again - now they have problems with memory, at least on my server);
- remove JIT warmup (JIT takes milliseconds, usually it doesn't affect the end results so much);
- pre-install required packages (like for Clojure, Julia and Haskell, however the latter could be the challenge and may require switching back to use Stack/Stackage);
- Haskell build for JSON could be improved by providing some flags and/or pre-installing the packages.


nuald:
@kostya I would recommend "Squash and merge" the PR to avoid polluting the history.

nuald:
Yes, the timing is from the docker run, and the slower results are not due to its overhead (I've verified locally), but rather up-to-dated tools (new JDK is slower, new Clang-based binaries are slower etc).

[Coursier ](https://get-coursier.io/) is required to fetch Maven artifacts (for JVM-based tests). It's used in Makefile-based builds as Maven and SBT do not use system-installed compilers thus making the tests unreliable.

Benchmarks are not copied into the image, they are assumed to be bind into /src (please see the README section regarding running the docker tests). It's intentional, so one can modify the code and verify it without rebuilding the image.

I'll address other comments shortly in the up-to-dated PR.

PR #201, opened 2019-11-08
Improve Racket bf2

philnguyen:
This pull request makes the improvements to the Racket version that are still reasonable, looking at other languages' versions

### DIFF
--- a/brainfuck2/bf.rkt
+++ b/brainfuck2/bf.rkt
@@ -1,78 +1,78 @@
 #lang racket/base
-(require racket/file racket/list racket/match)
+(require racket/file racket/list racket/match racket/cmdline
+         (rename-in racket/unsafe/ops
+                    [unsafe-vector-ref vector-ref]
+                    [unsafe-vector-set! vector-set!]
+                    [unsafe-fx+ +]))
 
-(struct op (op val) #:transparent)
-(struct tape (data pos) #:transparent)
+(define-syntax-rule (define-match-expander* k r) (define-match-expander k r r))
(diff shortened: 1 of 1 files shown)


nuald:
Could you please consider using `gvector`? As we try to have the consistent behaviour in the benchmarks, the tape is doubling its capacity when needed. It's either explicitly coded, or the default behaviour of the used containers. Looks like gvector also does doubling: https://github.com/racket/data/blob/master/data-lib/data/gvector.rkt#L41 Theoretically, it can improve performance of the code even more.

philnguyen:
`gvector` wraps around `vector`, which can slow things down. If we want doubling, we can just change that in the current implementation, can't we?

Anyways, would it be acceptable if I implement a tape as a "zipper" vector of 3 components: the left list, the element under focus, and the right list? I just tried it. It's faster for Racket.

Thanks ^_^

nuald:
If using the "zipper" vector would be an idiomatic approach (that would average developer would do), I think it's fine as soon as the core algorithm is the same. If not, you can always submit two versions - the reference implementation (that is close to others) and the optimized one. As Lisp-like languages are quite different from the imperative ones used in majority of the tests, I'm quite hesitant to decide what would be the "reference" for this particular use-case.

As for the original PR, yes, please use doubling (if you decide to stay with that version).

philnguyen:
I implemented vector doubling when full. Anything that significantly differs from all other implementation, I'll probably submit a different version.

Many thanks ^^

nuald:
Thanks, I'll update the numbers soon.

PR #203, opened 2019-11-09
Improve OCaml and F# versions

philnguyen:
This pull request fixes the interpretation of `loop` instruction in some functional versions to avoid repeatedly dispatching on the instruction, making it equivalent to imperative versions using `while`. This significantly speeds up F#.

It also turns on `-O` flag for F# and `-unsafe` for OCaml.

### DIFF
--- a/brainfuck2/bf.fs
+++ b/brainfuck2/bf.fs
@@ -51,9 +51,10 @@ let rec run program t =
       Console.Out.Flush()
       run ops t
   | Loop loop_code :: ops ->
-      if current t = 0 then run ops t
-      else
-        run loop_code t |> run program
+      let rec loop t =
+          if current t = 0 then run ops t
+          else run loop_code t |> loop
+      in loop t
 
 [<EntryPoint>]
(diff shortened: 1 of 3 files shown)


nuald:
Thanks for the good work, it's definitely going to be merged, but I'd like to clarify one issue first (not directly related to your change, however you may know the context better). F# version mentioned on the main page (v4.7) is mistakenly taken from .NET Core, however the used `fsharpc` is from the latest Mono (having v4.5).

As it's quite possible that F# is going to be provided by MS for .NET Standard and Core only  henceforth (https://devblogs.microsoft.com/dotnet/announcing-f-4-7/), do you think we need to update the tests to use .NET Core version? And if yes, could you please create a proper project file (as apparently MS team dropped support for using compilers directly: https://github.com/dotnet/cli/issues/7689)?

philnguyen:
Too bad I don't know .NET to create a project file :(

nuald:
No worries, I'll take a look later.

PR #207, opened 2019-12-03
improve Julia code and use Julia 1.3

wookay:
* use `julia-1.3.0`
* replace `brainfuck.jl` code from https://github.com/MikeInnes/BrainForth/blob/master/src/brainfuck.jl

### DIFF
--- a/base64/test.jl
+++ b/base64/test.jl
@@ -9,7 +9,7 @@ function main(tries)
 
   t = time()
   s = 0
-  for i in 1:tries
+  @inbounds for i in 1:tries
     str2 = base64encode(str)
     s += length(str2)
   end
@@ -20,7 +20,7 @@ function main(tries)
 
   t = time()
   s = 0
-  for i in 1:tries
+  @inbounds for i in 1:tries
     str3 = base64decode(str2)
     s += length(str3)
   end
--- a/base64/test.jl
+++ b/base64/test.jl
@@ -9,7 +9,7 @@ function main(tries)
 
   t = time()
   s = 0
-  @inbounds for i in 1:tries
+  for i in 1:tries
     str2 = base64encode(str)
     s += length(str2)
   end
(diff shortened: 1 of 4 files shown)


nuald:
`@inbounds` looks like a micro-optimization which we would like to avoid. However, it's acceptable to use the optimizing run/compile option (i.e. `--check-bounds=off`). Will it work for you?

Also, I guess (please confirm though as I'm not sure) that iterator or vector operations don't have the bound checks as unnecessary, and the proper (and more idiomatic) approach would be using those if possible. It would be nice to have the code following the proper approach, and as a side effect we won't need to disable bound checks.

wookay:
I'm not perfect but as I know that `getindex`, `length` have the bounds check during the loop.
so I put the `@inbounds` macro, and it got some improvements on my machine.
but I don't use the `--check-bounds=no` option directly in general.
I refer some docs about the bounds check in Julia, as a limitation of my explanation.
* docs
  - Checking bounds checking in Julia  https://vladium.com/post/20191016/checking-bounds-checking-in-julia/
  - Bounds checking https://docs.julialang.org/en/v1/devdocs/boundscheck/

nuald:
I've verified the bound check use-cases, and looks like `--check-bounds=no` gives better results in overall. I do realize that it could be not so commonly used, but it's aligned with the other tests we have (the code without micro-optimizations, and all optimizations are enforced by the compiler flags).

If you remove `@inbounds` usage, we can accept the PR (you may want to add `--check-bounds=no` to the required `run.sh`, but it's up to you - I can do it later by myself with the follow-up change with the results).

wookay:
ok. let it run.
I think that it's worth mentioning other benchmark site in julia discourse. 
- Julia programs now shown on benchmarks game website
  https://discourse.julialang.org/t/julia-programs-now-shown-on-benchmarks-game-website/17722

thank you.

nuald:
Thank you, I'll add `--check-bounds=no` and publish new results in the follow-up commit.

PR #215, opened 2019-12-13
The major update with the measurement deviances and a new energy metric.

nuald:
Main changes:
 - Automation has been added. The only reliable way I've found is using TCP sockets: tests signal the runner with their names and it starts measuring the metrics. Therefore the network stack has been added to all tests too.
- Energy metrics have been added. It's Linux-only measurements based on RAPL counters (those counters are CPU-specific, so could be used from any OS, but it would require additional code/tools). Unfortunately, my home server hardware is a little bit outdated as it has Sandybridge CPU that doesn't have DRAM energy consumption counters (http://web.eece.maine.edu/~vweaver/projects/rapl/rapl_support.html), so the tests are not accurate yet. Not sure when I'm going to update the server though.
- Standard deviation numbers have been added. We can rework the formatting later if you want.
- Python 2 support has been completely dropped as it hits its EOL just in few days. All Python tests are using Python 3 only.
 - TruffleRuby tests have been changed to use JVM version of TruffleRuby, the native one is way too unstable and pretty much slower for almost all tests.
- JSON file has been halved - Julia and TruffleRuby (both Native and JVM) occasionally crashed with the old one.

Additional notes:
 - Automation use 10 iterations. It may be not enough though as I've discovered that some slow tests which consume a lot of memory may have way too big deviations in the measurements. However, it's slow enough already, and I don't want to spend weeks on running tests :)
 - Libsocket has been added to be used for C/C++ and Chez Scheme. While the former could use POSIX sockets from libc directly, the latter would require way too much FFI code to embed it (and I couldn't find easy to use Chez Scheme libraries for simple networking). Libsocket is easy to use from FFI, and as soon as it adds nice API for C/C++ (and it's idiomatic, for example, it has RAII for C++), I've changed C/C++ code to use it too. It would be nice to use it in Swift (the current code utilizing POSIX socket is quite ugly), but I didn't want to create a special project for Swift test (as it looks like it's not possible to transparently use C libraries in it).
 - jq tests are not using jq's default C-based frontend utility `jq` anymore, and have been switched to Rust-based frontend. Internally it does the same - compile and execute jq program utilizing `libjq`, and has the same performance, but it fits the test framework we're using (as original `jq` can't send network requests). 

Future work:
 - The tests are not accurate yet as the measurements stop when the process is unloaded from the OS and it may take time, especially for GC-based languages. I've seen that the numbers could differ for 0.4 seconds for base64 tests. It's quite easy to fix - tests could signal the runner when to stop the measurements with the additional TCP request, but it would require even more changes in this already big PR.


kostya:
Looks ok. not quite understand why socket needed, it seems it useful for jvm, to check after warming, but why for others?

nuald:
It's not only for JVM, warming up is required for .NET and Julia too (frankly, I think that other languages may need it too). Another problem that has been solved is avoiding timing of in-place compilation: for example, Swift compilation takes few seconds before actually running the code, so the measurements for it were wrong.

You're right that we don't need it for native binaries, but please let me notice that loading/unloading processes may take some time too (for example, json_go binary is 3.5 Mb, and it definitely takes few milliseconds to load/unload it).

As I've noted, even the new implementation is not accurate enough. However if the tests notify the runner when to start/stop, we can improve the accuracy even more. And the simplest approach for those notifications is using sockets. I've tried pipes too, but it didn't work so well. Surely, it made the code more complex, but unfortunately I couldn't find another way how to automate the tests and get the accurate results at the same time.

beached:
The libsocket++ library is not compatible with macos.  
`C/inet/libinetsocket.c:1013:46: error: use of undeclared identifier 'SIOCGIFINDEX'
            if (-1 == check_error(ioctl(sfd, SIOCGIFINDEX, &interface))) {`
`C/inet/libinetsocket.c:1020:43: error: no member named 'ifr_ifindex' in 'struct ifreq'
            mreq4.imr_ifindex = interface.ifr_ifindex;`
`C/inet/libinetsocket.c:1051:46: error: use of undeclared identifier 'SIOCGIFINDEX'
            if (-1 == check_error(ioctl(sfd, SIOCGIFINDEX, &interface))) {`
`C/inet/libinetsocket.c:1058:48: error: no member named 'ifr_ifindex' in 'struct ifreq'
            mreq6.ipv6mr_interface = interface.ifr_ifindex;`
`C/inet/libinetsocket.c:1061:61: error: use of undeclared identifier
      'IPV6_ADD_MEMBERSHIP'
        if (-1 == check_error(setsockopt(sfd, IPPROTO_IPV6, IPV6_ADD_MEMBERSHIP,`

alex-slesarev:
@beached Unfortunately, macOS is not officially supported. libsocket is not the only incompatibility here, energy and memory readings are Linux-specific too. If you would like to have C/C++ examples working on macOS and willing to contribute, I'd suggest two options:
- Find another library with the simple API. POSIX API requires a lot of definitions provided by FFI (in particular, we had a problem with Chez Scheme - it doesn't have networking embedded, but rather utilize external C libraries). There are other libraries like plibsys, but the latter require a lot of FFI definition too.
- Write static library within the test suite and use it from C, C++ and Chez Scheme. We won't need a full networking API, something similar to `notify` call could be enough (that would connect to the server, send a text and close the connection).

Otherwise, I'd recommend you to open a ticket and we may eventually address it.

nuald:
@beached Unfortunately, macOS is not officially supported. libsocket is not the only incompatibility here, energy and memory readings are Linux-specific too. If you would like to have C/C++ examples working on macOS and willing to contribute, I'd suggest two options:

- Find another library with the simple API. POSIX API requires a lot of definitions provided by FFI (in particular, we had a problem with Chez Scheme - it doesn't have networking embedded, but rather utilize external C libraries). There are other libraries like plibsys, but the latter require a lot of FFI definition too.
- Write static library within the test suite and use it from C, C++ and Chez Scheme. We won't need a full networking API, something similar to `notify` call could be enough (that would connect to the server, send a text and close the connection).

Otherwise, I'd recommend you to open a ticket and we may eventually address it.

beached:
I was able to get libsocket++ going(ish) on macos.  It was detecting macos as linux.

nuald:
@beached Please don't worry about libsocket anymore, it has been replaced by the custom library (the second item in the suggested list) in #229. It should be compatible with macOS, I've verified few tests (not sure all C/C++ examples would work on macOS though, but if you have any issues, please open a ticket and we may consider fixing it).

beached:
Oh nice,  I have been running the c++ tests mostly to compare my own library and they generally do work.

PR #217, opened 2019-12-16
Add brainfuck2 for Vala

gavr123456789:
Vala language: https://wiki.gnome.org/Projects/Vala  
Other benchmarks: https://github.com/arrufat/vala-benchmarks   
Just copied the C# version, changing the name of the timer, output and data structures, Vala has exactly the same syntax.

### DIFF
--- a/brainfuck2/bf.vala
+++ b/brainfuck2/bf.vala
@@ -1,3 +1,40 @@
+#!/usr/bin/vala
+//Compile $ valac bf.vala --disable-assert -X -O3 --pkg gee-0.8 --pkg gio-2.0 -o bin_vala_gcc
+//Run like script $ ./bf.vala --disable-assert -X -O3 --pkg gee-0.8 --pkg gio-2.0 --run-args bench.b
+void print_vala_version() {
+    string vala_version="",error="";
+    try {
+        Process.spawn_command_line_sync ("valac --version", out vala_version,out error);
+        if (error.length > 1) print (error);
+        else print (vala_version.split(" ")[1] + "\n");
+    } catch (SpawnError e) {print (@"Error: $(e.message)\n");}
+}
+
+void send_language_name () {
+    var host = "localhost:9001";
+    try {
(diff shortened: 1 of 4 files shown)


nuald:
There have been changes recently in the testing framework, and now all tests are notifying the runner when to start the measurement. Could you please change the code accordingly (it should send "Vala" message to localhost:9001 via TCP socket, and close the socket before running the actual test + fail gracefully in the case the server is not available)? Looks like it shouldn't be too challenging (https://wiki.gnome.org/Projects/Vala/GIONetworkingSample), however if it's a problem, I can deal with it later.

Also, do you know how to detect the Vala version programmatically? "valac --version" includes the name, not the version only. We try to avoid parsing the CLI output, and rather use other means to get the version as-is, and if you know the way, I'd be grateful if you share the knowledge. But again, I can research it by myself, no worries here.

gavr123456789:
Thanks for the quick response. I have never dealt with networks, so I did something from this example, I think you will not be difficult to convert it into what you need.  
 
About obtaining version of Vala without language name, i think this problem will emerge with many languages, itself gcc for example prints 4 lines of text on request version of. I suggest if you don't want to parse the name on the server side, why not do it on the side of the test itself. I added a function which safely outputs the version of the first line of test.  
  
Another idea is to implement server-side parsing, but to do it through a package manager version request, then it will be universal for all languages.

nuald:
Thanks, I’ll use it as a basis and include you into the final PR with numbers to review.

If you don’t mind, we’ll use not the latest version of the Vala compiler, but rather the version included in Debian testing (looks like it’s just one minor version behind). I’d rather avoid installing from sources as Docker may have issues with it, and their PPA is even more behind than Debian.

gavr123456789:
Okay great. Of course I do not mind, not the latest version of Vala is unlikely to affect performance due to compile in C. But just for the record, I dont know whether Debian can connect PPA as Ubuntu, but in this [PPA](https://launchpad.net/~vala-team/+archive/ubuntu/next) vala maintainer puts the latest releases.

PR #219, opened 2019-12-17
Follow up for #217.

nuald:
The Vala code changes:
 - added GCC/Clang tests detection;
 - added flush() call for the consistency with the other tests.

Please ignore the additional Dockerfile changes not related to Vala tests - as the image has to be rebuilt almost from the scratch, I've updated other languages too, but the new results will be published later during the scheduled update.


nuald:
@gavr123456789 I can't officially assign you for the review, so please just take a look and let me know if the changes are fine for you.

gavr123456789:
It's strange, but on my machine I consistently get better results on vala than C++, when testing with bench.b.  Anyway all ok.
image

nuald:
@gavr123456789 - looks like you're verifying on the old codebase as the current C++ version use libsocket++. Vala compilation flag is missing "-D GCC_TEST" too. Networking does add some overhead, plus looks like you have faster and more modern CPU that may have better branch predictions comparing with mine.

PR #221, opened 2019-12-22
Add a version of Java matmul that uses ND4J.

wmeddie:
Similar to Python and numpy this version of matmul uses the ND4J to do the matrix operations.

The library is added by the nd4j-build maven project which creates an uberjar with nd4j as well as the native binaries for linux-x86_64 and AVX2.

The performance on my machine was in line with numpy, which is significantly faster than the pure Java version.  It might be worth increasing the benchmark from 1500 to 5000 or so to be able to see the difference with these optimized versions.

The output from my computer comparing numpy, nd4j and pure java are below:
 
```
(venv) egonzalez@DESKTOP:/mnt/e/src/github/benchmarks-linux/matmul$ ../xtime.rb python matmul-numpy.py 1500
checking for rapl-info... no
Python NumPy
-143.5001666666568
0.17 s, 79.3 Mb
(venv) egonzalez@DESKTOP:/mnt/e/src/github/benchmarks-linux/matmul$ ../xtime.rb java -cp nd4j-build/target/nd4j-build-1.0-jar-with-dependencies.jar:. matmulnd4j 1500
checking for rapl-info... no
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
JIT warming up: -99.49933733840005
Java
-143.50016666665678
time: 0.196s
0.22 s, 242.2 Mb
(venv) egonzalez@DESKTOP:/mnt/e/src/github/benchmarks-linux/matmul$ ../xtime.rb java matmul 1500
checking for rapl-info... no
JIT warming up: -99.49933733840005
Java
-143.5001666666568
time: 3.292s
3.31 s, 120.1 Mb
```


matmul/matmulnd4j.java
```diff
@@ -0,0 +1,41 @@
...
+		int n = 100;
+		if (args.length >= 1) n = Integer.parseInt(args[0]);
+		n = n / 2 * 2;
+		
+		INDArray t =  Nd4j.matmul(matgen(500), matgen(500));
+		System.out.println("JIT warming up: " + t.getDouble(1, 1));
+		
+		try (Socket socket = new Socket("localhost", 9001);
+			OutputStream out = socket.getOutputStream()) {
+			out.write("Java".getBytes("UTF-8"));
```
nuald: Please use a test name (that would be presented in the result table), I guess it could be "Java ND4J"

wmeddie:
Switched to the coursier-based build.  I'm not sure how to specify a classifier so instead it just builds using nd4j-native-platform, which works and is the same but it also includes the native binaries for windows, android, and mac, etc.

wmeddie:
Figured out how classifiers work.  So now the build only uses linux-x86_64-avx and produces a slightly more reasonable 71MB uberjar instead of a 400MB one.

nuald:
Thanks, the results will be published soon.

nuald:
I've published the results, however I had to use the generic CPU classifier as the test server doesn't support AVX2 (as it's Sandy Bridge CPU). I've added a note into Maven.list to use AVX2 or better as soon as the server is upgraded.

wmeddie:
Fair enough.  I wish we could do that automatically somehow.  Maybe in the future.

PR #223, opened 2020-01-02
add arraymancer, add opt:speed to matmul.nim compilation

machineko:
Add arraymancer results (numpy like nim library)
Add  --opt:speed flag to base nim matmul results


nuald:
A follow-up will be published a little bit later. Meanwhile, just out of curiosity, why did you use a special gc flag? Does it affect performance or have something to do with arraymancer?

machineko:
It was pure performance i saw it was used in other benchmarks by nim team members so i also use it here.
For 1500x1500 array  it probably doesnt matter but for some heavier tasks it should boost performacne.

PR #225, opened 2020-01-04
json: Use JSON3 in test.jl

cmcaine:
The `JSON` package used isn't part of the standard library, so maybe it's fair
to replace it with the newer package by the same author?

On my machine, `JSON3` is over twice as fast and uses much less memory.


nuald:
Looks like JSON3 is considered as experimental, and could be eventually merged to the upstream JSON: https://discourse.julialang.org/t/ann-json3-jl-yet-another-json-package-for-julia/25625 On the other hand, JSON.jl occasionally fails on my server due to the memory consumption and with new Julia release it's happening even more often thus preventing me from getting benchmark results. So, I'm going to merge it, but eventually we may reconsider it and have 2 separate tests - for JSON and JSON3.

PR #226, opened 2020-01-09
The benchmarks accuracy has been improved.

nuald:
The major changes:
 - The communication protocol between tests and runner has been improved to include start/stop notifications and PID of the running process (measuring parent+child processes memory were giving slightly inaccurate results especially for languages like Elixir having auxiliary processes running alongside the main ones);
 - File system I/O has been moved out from measurement boundaries (docker has FS issues giving inconsistent results);
 - JSON blob has been moved to `/tmp` as Docker has issues with reading big files from binded FS (tests were failing as docker provided corrupted file content).

With the improved accuracy I've noticed that there are inconsistencies caused by console I/O (doesn't matter for slow languages, but can affect faster ones with +/- 1ms). I guess the improvement regarding that could be included in the next scheduled update.


kostya:
looks ok!
Rust in son first place because of /tmp change, or update lang/lib?

nuald:
I've researched the "GDC fast" performance, but couldn't find the actual underlying reason for the performance differences after the code changes:
- original code uses parseJSONFile and has the avg 15ms execution time;
- new code uses parseJSON and has the avg 19ms execution time.

My guess is that the GDC compiler has some optimizations for mmap-ed regions (opposite to usual D strings) as parseJSON doesn't actually read the file into the newly allocated memory, but map the file: https://github.com/mleise/fast/blob/master/source/fast/json.d#L1479

However, it's rather GDC optimization, the DMD compiler shows the same performance for both methods. Considering that DMD is the official reference compiler, I think it's worth to stick with it (even if its results a little bit slower), thus I've changed all additional tests to use it (it's actually just 2 - this one + matmul lubeck).

kostya:
as i know noone using dmd in production, so i think better to return gdc or add gdc as it faster.

nuald:
It's a good point and thank you for forcing me to research it a little bit further. I've found the culprit - the memory reallocation due to SSE alignment required (during parsing JSON file they don't need to do any reallocation as they utilize mmap functionality). The code has been fixed and the numbers updated.

PR #227, opened 2020-01-19
Added bf2 C implementation

p7g:
I wrote an bf interpreter in C, in the same way as the other language implementations work.

Regarding the criteria in the overview of the readme:
- I would say I'm a pretty average C programmer, and the only "optimizations" I did were finding and removing memory leaks with valgrind.
- Algorithms are the same as in the other languages.
- Array lists were used for the tape and op lists, which should be pretty common in C programs

It seems to be pretty fast.

### DIFF
--- a/brainfuck2/bf.c
+++ b/brainfuck2/bf.c
@@ -288,6 +288,9 @@ int main(int argc, char *argv[])
 	fsize = ftell(f);
 	fseek(f, 0, SEEK_SET);
 
+	/* use unbuffered stdout */
+	setbuf(stdout, NULL);
+
 	code = malloc(sizeof(char) * fsize + 1);
 	fread(code, 1, fsize, f);
 	fclose(f);
@@ -303,7 +306,6 @@ int main(int argc, char *argv[])
 	parse(&it, &ops);
 
 	eval(&ops, &tape);
-	fflush(stdout);
 
 	free(code);
 	tape_free(tape);
--- a/brainfuck2/bf.c
+++ b/brainfuck2/bf.c
@@ -79,18 +79,9 @@ void op_list_free(struct op_list *list)
 
 void op_list_grow(struct op_list *list)
 {
-	struct op *new_list;
-	int i;
-
 	/* double the capacity */
 	list->cap <<= 1;
-	new_list = malloc(sizeof(struct op) * list->cap);
-
-	for (i = 0; i < list->len; i += 1)
-		new_list[i] = list->ops[i];
-
-	free(list->ops);
-	list->ops = new_list;
+	list->ops = realloc(list->ops, sizeof(struct op) * list->cap);
 }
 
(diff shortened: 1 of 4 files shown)


brainfuck2/bf.c
```diff
@@ -0,0 +1,317 @@
...
+}
+
+void op_list_grow(struct op_list *list)
+{
+	struct op *new_list;
+	int i;
+
+	/* double the capacity */
+	list->cap <<= 1;
+	new_list = malloc(sizeof(struct op) * list->cap);
```
nuald: Please use realloc - it aligns better with the other tests and potentially faster.

brainfuck2/bf.c
```diff
@@ -0,0 +1,317 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <libsocket/libinetsocket.h>
+
+#define OP_LIST_DEFAULT_CAPACITY 1 << 5
+#define TAPE_DEFAULT_CAPACITY 1 << 5
```
nuald: The default capacity of tape is 1 element in other tests.

brainfuck2/bf.c
```diff
@@ -0,0 +1,317 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <libsocket/libinetsocket.h>
+
+#define OP_LIST_DEFAULT_CAPACITY 1 << 5
```
nuald: The default operation list is empty.

p7g:
Thanks for the feedback!  I didn't realize that most dynamic array implementations have no initial capacity, so thanks for pointing that out. In my investigation I noticed that java's ArrayList might have an initial capacity of 10 [here][1] (not sure if [this documentation][2] applies), but that's definitely an exception, not the rule.

With regards to the unbuffered output, I'm open to replacing that with an `fflush(stdout)` after printing the char if that would even the playing field a bit more.

I'll fix all of this up later today.

[1]: https://github.com/kostya/benchmarks/blob/master/brainfuck2/bf.java#L75
[2]: https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html#ArrayList--

nuald:
Interesting, I wasn't aware of initial capacity of 10. I guess it's some internal optimization and other languages may have it too, but it doesn't necessary improves the performance. I've tried the Rust example to use Vec::with_capacity(10), but actually got worse results: from 3.37 ± 00.04 to 3.41 ± 00.06. JVM may have some JIT optimizations to not allocate 10 elements if they're not going to be used, but obviously it's not applicable to other tests and I guess we're going to stick to empty lists for now.

p7g:
I've gone and fixed all of the things you pointed out. I think the `realloc` change made quite a difference in performance.

Please let me know if you find any other issues.

PR #231, opened 2020-01-31
DAW JSON Link C++ JSON parser test

beached:


beached:
I would love to add my JSON library.  It performs quite well and takes an approach like Serde Typed but in C++

json/build.sh
```diff
@@ -90,3 +90,14 @@ go build -o json_iter_go test_jsoniter.go
 if [ ! -f /tmp/1.json ]; then
   ruby generate_json.rb
 fi
+
+if [ ! -d ./daw_json_link ]; then
```
nuald: I'd like to centralize the vendor libraries usage here (as they seem to semi-connected). Git submodules could be an overkill, but using one directory could fit better. Something like that:
```
if [ ! -d ./daw_json_link_all ]; then
  git clone --depth 1 https://github.com/beached/daw_json_link.git ./daw_json_link_all/daw_json_link
  git clone --depth 1 https://github.com/beached/header_libraries.git ./daw_json_link_all/header_libraries
  git clone --depth 1 https://github.com/beached/utf_range.git ./daw_json_link_all/utf_range
fi
```

json/test_dawjsonlink.cpp
```diff
@@ -0,0 +1,102 @@
...
+		using type = json_member_list<json_array<"coordinates", coordinate_t>>;
+#else
+		constexpr inline static char const coordinates[] = "coordinates";
+		using type = json_member_list<json_array<coordinates, coordinate_t>>;
+#endif
+	};
+} // namespace daw::json
+
+int main( int argc, char *argv[] ) {
+	std::ios_base::sync_with_stdio( false );
```
nuald: Is it needed?

json/test_dawjsonlink.cpp
```diff
@@ -0,0 +1,102 @@
...
+	std::string_view fpath = "/tmp/1.json";
+	if( argc > 1 ) {
+		fpath = argv[1];
+	}
+	auto const text = daw::filesystem::memory_mapped_file_t<>( fpath );
```
nuald: Other tests load the file into the memory. I couldn't say will it affect performance or not, but for the consistency it would be better to do the same (please note that the tests run in Docker, and it may affects the file system operations).

json/test_dawjsonlink.cpp
```diff
@@ -0,0 +1,102 @@
...
+	auto const json_sv = std::string_view( text.data( ), text.size( ) );
+	double x = 0, y = 0, z = 0;
+	int len = 0;
+
+	using range_t = daw::json::json_array_range<coordinate_t, true>;
```
nuald: Please put this line and below after the notification. While it may have the minimal impact, we'd prefer all the JSON-related operations (except loading the file into the memory as it's not about JSON library performance, but rather FS performance) be in the measurement scope (between notifications).

json/build.sh
```diff
@@ -90,3 +90,14 @@ go build -o json_iter_go test_jsoniter.go
...
+  git clone --depth 1 https://github.com/beached/daw_json_link.git
+fi
+if [ ! -d ./header_libraries ]; then
+  git clone --depth 1 https://github.com/beached/header_libraries.git
+fi
+if [ ! -d ./utf_range ]; then
+  git clone --depth 1 https://github.com/beached/utf_range.git
+fi
+g++ -std=c++17 -O3 -march=native -L /usr/local/lib -I /usr/local/include -I./daw_json_link/include -I./header_libraries/include -I./utf_range/include test_dawjsonlink.cpp -o json_dawjsonlink_cpp -I../common/libnotify -L../common/libnotify -lnotify
```
nuald: `-march=native` isn't used by other GCC tests. For the consistency, I'd prefer we avoid using it (or add it to all other tests). I've tried to use it before, and discovered that it doesn't affect the performance so much (or could even lead to controversial results). It could be GCC-related though (https://lemire.me/blog/2018/07/25/it-is-more-complicated-than-i-thought-mtune-march-in-gcc/).

json/build.sh
beached: Did this change and added to .gitignore

json/test_dawjsonlink.cpp
beached: Wasn't just habbit with benching involving iostreams

json/test_dawjsonlink.cpp
beached: Used same readfile as rapid in latest rev

json/test_dawjsonlink.cpp
beached: the next line did happen to be significant, move all of it below notify

json/build.sh
beached: removed -march=native

beached:
Updated PR with requested modifications

nuald:
The follow-up with the numbers will be provided in a few days - I'm in the middle of the maintenance update that takes few days to finish.

beached:
Thanks.  On my machines, I am doing well.  But different machines and circumstances can have an impact.

PR #234, opened 2020-02-09
Use Unmarshal for Go

proyb6:
Remove the conversion to string part, it can read json file into struct, result in less code.

Original:
Memory: <411MB
Time: 1.95s

New Code:
Memory: 215MB
Time: 1.7s

### DIFF
--- a/json/test.go
+++ b/json/test.go
@@ -7,7 +7,6 @@ import (
 	"net"
 	"os"
 	"runtime"
-	"strings"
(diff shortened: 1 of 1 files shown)


nuald:
Looks good, but could you please apply gofmt as looks like there are inconsistent indentation in the changed code.

proyb6:
@nuald Thanks, formatted.

PR #240, opened 2020-03-10
brainfuck.jl: don't reimplement UInt8 rollover

cmcaine:
If you're working with bytes it is idiomatic to add and subtract UInts 
directly rather than promoting to Int, doing some branches and then casting
down to UInt again.

Credit to Harmen Stoppels on julialang.slack.com

I also symlinked:

   ln -sr brainfuck/brainfuck.jl brainfuck2/bf.jl

They're functionally identical, so they might as well be the same file.


nuald:
Thanks, the up-to-dated numbers will be provided within the maintenance update in a couple of weeks. If you like to have the numbers now, please let me know.

cmcaine:
You're welcome. I'm in no hurry :)

This is ~4% faster on my machine for the first benchmark. Harmen got ~10-15% speedup.

Issue #255, opened 2020-06-02
ConcurrentGarbageCollection for .NET Core


OpenSourceAnarchist:
Similar to #251, I noticed that the (questionable) Computer Language Benchmarks Game enables _both_ ServerGarbageCollection and ConcurrentGarbageCollection, as shown at https://benchmarksgame-team.pages.debian.net/benchmarksgame/measurements/csharpcore.html

I don't use any .NET languages, but any standard optimizations should be used for all represented languages.

nuald:
ConcurrentGarbageCollection is enabled by default (https://docs.microsoft.com/en-us/dotnet/core/run-time-config/garbage-collector), but I've set it explicitly in the new upcoming major update.

OpenSourceAnarchist:
I wasn't sure if setting ServerGarbageCollection to true would set ConcurrentGarbageCollection to false since Microsoft recommends concurrent gc for workstations, not for servers. Regardless, thanks for making it explicit and I look forward to the new update!

nuald:
Addressed in #257.

Issue #248, opened 2020-04-16
Implement better testing framework


nuald:
Right now the JIT and lazy initialization may interfere with the results, and to avoid it we can implement the following:

- run the tested code in the loop and as soon as the previous result is faster than the current, stop the loop (surely, it could be a fluctuation caused by the external reason, but running **all** tests X times would take forever to execute)
- extending the test runner communication protocol to support several notifications from within one test (and write to the results.log file only the best result).

Unfortunately, it requires changes in all tests.

eregon:
That sounds good, running the actual benchmark workload 2+ times seems better than a warmup with a different workflow.

Another way as I suggested in #246 is to not warmup, just avoid counting startup. Lazy initialization that takes 0.6s seems excessive and arguably an issue of the library used.

nuald:
JIT warming up has been removed, and lazy initialization issues are addressed (not finished yet though) with the testing of the code to work properly before the benchmark.

The proposed idea has been tested, and dismissed. While the 2nd or 3rd iteration could be faster, it's not always happening as GC could affect the results quite drastically. Fine tuning is required, and it's out of scope of that project.

PR #257, opened 2020-06-15
Major update of the build infrastructure

nuald:
I haven't started that change as the build infrastructure changes, but as I started to work on code changes (adding internal tests), using the commands manually became a little bit annoying, so I fixed it.

Major changes are:
 - replaced clean.sh/build.sh/run.sh with the Makefiles with the proper dependencies;
 - changed .NET Core GC settings (#255);
 - removed JIT warmup and instead added internal tests (#248). JIT warmup gave controversial results (even if the code could execute faster, GC could diminish the results drastically). The only real advantage of JIT warmup was the initialization of the lazy code, and it's now addressed by additional  tests before running a benchmark (not implemented yet for all tests though);
 - added entry points to all tests (as some scripting languages had the entry code through the whole script). While it's mostly the code style issue, it makes the refactoring easier (as everything encapsulated in one place and have the proper indentation);
 - removed old brainfuck (or rather renamed bf2->bf) to avoid code duplication and additional maintenance (unfortunately, it made the github diff quite huge, however it was quite minor change);
 - some reformatting (mostly automatic) has been applied as I've added new tools for checking style and autoformat (not for all languages though).

Pending changes:
 - add internal tests for all benchmarks (besides dealing with lazy initialization it helps to find the minor coding issues, i.e.  some matmul tests instead of allocating memory for 2 matrices, allocated to just one and reused it for another, thus having unfair advantage);
 - add some formatters and code style checkers (optional, but helps to find the code that is not idiomatic).


kostya:
Looks ok.
* Question why C++ brainfuck much faster? related to gcc update?
* Btw, in other benchmarks i use output time to stderr instead of socket (with some markers), it much easier to do, and also xtime.rb and the programs, runs easier independent of each other.

nuald:
GCC 10 indeed improved the performance of the C++ bf test, I've verified with g++9 (1.83s) and g++10 (1.03s).

As for the measuring the timing, yes - it's simpler to utilize stderr. However, there are several factors here:

- The timers in various languages could be not accurate enough, the most obvious example is reduced precision in JS engines introduced after infamous timing attacks like Meltdown and Spectre: https://developer.mozilla.org/en-US/docs/Web/API/Performance/now#Reduced_time_precision
 - As we need to grab additional metrics here (like memory and energy consumption), it's easier to have one unified logic (in the test runner) comparing with modifying all tests to provide those metrics. The good example is memory consumption - we'd like to get the maximum allocated memory, so we need to measure it during the test execution (we can't measure it only after we finish as GC could clear it already), and it could be not trivial at all.

PR #260, opened 2020-07-12
Rust base64: use the radix64 library

adnan338:
Performance boost upto 4 times or more depending on your CPU. My time went from 1.4s to 0.24s.


nuald:
Please create this test as the separate one from base64 (not replacing it). You could use the same method as in JSON-tests and provide few binaries within one Cargo project. I guess the main reason why I would like to have base64 to be tested too is because it supports `no_std` and don't have dependencies (therefore more friendly to the embedded projects).

kostya:
is this 4 times increase related to using buffer, instead of allocation?

adnan338:
> is this 4 times increase related to using buffer, instead of allocation?

Buffer certainly helps, but even without using buffer, it's faster due to SIMD aware optimizations.

adnan338:
> Please create this test as the separate one from base64 (not replacing it). You could use the same method as in JSON-tests and provide few binaries within one Cargo project. I guess the main reason why I would like to have base64 to be tested too is because it supports `no_std` and don't have dependencies (therefore more friendly to the embedded projects).

Restored `base64` with the previous code, removed manual buffering. 

```
adnan338@pop-os:~/proj/benchmarks/base64/base64.rs$ 
adnan338@pop-os:~/proj/benchmarks/base64/base64.rs$ cargo build --release
   Compiling base64_rs v0.0.1 (/home/adnan338/proj/benchmarks/base64/base64.rs)
    Finished release [optimized] target(s) in 2.24s
adnan338@pop-os:~/proj/benchmarks/base64/base64.rs$ time ./target/release/base64 
encode aaaa... to YWFh...: 1431666688, 0.53376377
decode YWFh... to aaaa...: 1073741824, 0.82510775

real	0m1.361s
user	0m1.361s
sys	0m0.000s
adnan338@pop-os:~/proj/benchmarks/base64/base64.rs$ time ./target/release/radix64 
encode aaaa... to VTC_...: 1431658496, 147
decode VTC_... to aaaa...: 1073741824, 104

real	0m0.254s
user	0m0.250s
sys	0m0.004s

```

PR #282, opened 2020-10-31
Update test_boost_json.cpp

beached:
This shows the Boost JSON in a better light and uses safer at( ) instead of operator[] that will insert if the member doesn't exist.  At throws.  It also uses a more idiomatic, for read only JSON, use of memory_resource similar to RapidJSON's pooled allocator.

### DIFF
--- a/json/test_boost_json.cpp
+++ b/json/test_boost_json.cpp
@@ -37,15 +37,15 @@ string read_file(const string& filename) {
 coordinate_t calc(const string& text) {
   auto x = 0.0, y = 0.0, z = 0.0;
   auto len = 0;
-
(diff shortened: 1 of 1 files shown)


nuald:
If you want safer operations, I'd recommend to change `get_` to `as_` too (i.e. use `as_double` instead of `get_double`). I deliberately used unsafe []-operator because we know ahead that the original JSON have the proper format, however in the real life I guess we want to use the safer operations as usually JSON values come across the trust boundaries.

beached:
At is also faster than operator[] when I was checking.  It has a const overload too, so if the object is const it would still work.

PR #283, opened 2020-11-01
Lua perf boost

ExtReMLapin:
Compatible with LuaJIT (using FFI structs), Lua 5.1 and Lua 5.4

Performance change compared to [brainfuck interpreter](https://github.com/kostya/benchmarks/blob/master/brainfuck/bf.lua)

LuaJIT :

	mandelbrot -> ran in 53.81s -> ran in 2.183s

	bench -> ran in 8.25s -> ran in 0.183s

**2364.96%** and **4408.2%** perf boost

Lua 51 : 

	bench -> ran in 98.926s -> ran in 14.042s

**604.5% perf boost**

Lua 54 :

	bench -> ran in 51.345s -> ran in 4.858s

**956.92%** perf boost

LuaJIT goes BRRRRRRR

### DIFF
--- a/brainfuck/bf.lua
+++ b/brainfuck/bf.lua
@@ -1,111 +1,113 @@
-local INC = 0
-local MOVE = 1
-local PRINT = 3
-local LOOP = 4
+local specialInstruction = {
+	["."] = "w(data[i])",
+	[","] = "data[i]=r()",
+	["["] = "while data[i]~=0 do ",
+	["]"] = "end "
+}
 
-local function Tape()
-  local pos = 1;
-  local data = {0};
+local artithmeticsIns = {
+	["+"] = "data[i] = data[i]+",
+	["-"] = "data[i] = data[i]-",
+	[">"] = "i=i+",
+	["<"] = "i=i-"
+}
 
-  local get = function()
-    return data[pos];
-  end
+local brainfuck = function(s)
+	s = s:gsub("[^%+%-<>%.,%[%]]+", "") -- remove new lines
+	local instList = {}
+	local slen = #s
+	local i = 2
+	local lastInstruction = s:sub(1, 1)
+	local arithmeticsCount = 0
 
-  local inc = function(x)
-    data[pos] = data[pos] + x;
-  end
+	if (artithmeticsIns[lastInstruction]) then
+		arithmeticsCount = 1
+	end
 
-  local move = function(x)
-    local length = #data;
-    pos = pos + x;
-    for i = length + 1, pos do
-      data[i] = 0;
-    end
-  end
+	while (i <= slen) do
+		local curInst = s:sub(i, i)
 
-  return {
-    get = get;
-    inc = inc;
-    move = move;
-  };
+		if curInst == lastInstruction then
+			if artithmeticsIns[curInst] then
+				arithmeticsCount = arithmeticsCount + 1
+			else
+				table.insert(instList, specialInstruction[curInst])
+			end
+		else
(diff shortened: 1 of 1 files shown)


kostya:
You completely change algorithm, this is not accepted. We can do similar in other languages and have boost too.

ExtReMLapin:
Alright alright let me do another pull request with minimum code modifications

kostya:
we can use different implementation details and tunes, but algorithm should be the same, to be comparable between languages.

ExtReMLapin:
Could this be accepted @kostya ?


```lua
local INC = 0
local MOVE = 1
local PRINT = 3
local LOOP = 4
local isJIT = type(jit) == "table"

local function Brainfuck(text)
	local pos = 1
	local data

	if isJIT then
		data = require("ffi").new("int[30000]") -- cf: wikipedia
	else
		data = {}
		local i = 0

		while (i < 30000) do
			data[i] = 0
			i = i + 1
		end
	end

	local insData = {
		["+"] = {INC, 1},
		["-"] = {INC, -1},
		[">"] = {MOVE, 1},
		["<"] = {MOVE, -1}
	}

	local function parse(source, i)
		local res = {}
		local len = source:len()

		while i <= len do
			local c = source:sub(i, i)
			local insPtr = insData[c]

			if insPtr then
				local count = 1

				while i <= len and source:sub(i + 1, i + 1) == c do
					count = count + 1
					i = i + 1
				end

				table.insert(res, {insPtr[1], insPtr[2] * count})
			elseif c == "." then
				table.insert(res, {PRINT})
			elseif c == "[" then
				local loop_code
				loop_code, i = parse(source, i + 1)
				table.insert(res, {LOOP, loop_code})
			elseif c == "]" then
				break
			end

			i = i + 1
		end

		return res, i
	end

	local function _run(program)
		for i = 1, #program do
			local op = program[i]
			local operator = op[1]

			if operator == INC then
				data[pos] = data[pos] + op[2]
			elseif operator == MOVE then
				pos = pos + op[2]
			elseif operator == LOOP then
				while data[pos] > 0 do
					_run(op[2], nil)
				end
			elseif operator == PRINT then
				io.write(string.char(data[pos]))
				io.flush()
			end
		end
	end

	local function run()
		_run(parse(text, 1))
	end

	return {
		run = run
	}
end
```

kostya:
You remove Tape abstraction, are you think this is good for real tasks usage? Should not language inline it by itself? Also prefer to use autogrowing array, as all languages use.

ExtReMLapin:
I removed Tape abstraction for personal readability.

The wikipedia article states it needs an array of 30k, shouldn't all language use it ?

kostya:
We decide just use everywhere idiomatic array storage.

nuald:
@ExtReMLapin The purpose of the bf benchmark is testing the performance of the idiomatic constructions, it could be any other general-purpose algorithm. However, I totally understand that you may want to have better, optimized version for the reference (in this particular example, FFI makes sense as Lua is intended to be embedded and communicate a lot with the native code). 

Therefore, my recommendation would be not changing the reference implementation (that should have the similar algorithm and the similar list/array operations), but add a new one and mark it as "Lua (FFI)" (I'm not sure FFI is a reason for the performance increase though, I'd leave the proper naming to you). You could see that we have the same with Haskell (reference implementation and MArray-based). If you have troubles adding it into the toolchain, just please submit a new file, and I'll add it to all required places (just let me know do you need it for Lua and/or LuaJIT).

Also, if you see some obvious mistakes in the reference implementation, you're welcome to address those unless algorithm changes are required.

PR #296, opened 2021-02-24
Add Racket BF variant that uses syntax objects and `eval`

philnguyen:
This pull request adds a Racket variant for the `bf` benchmark that parses bf syntax directly into Racket syntax, omitting the explicit interpretation loop and dynamic dispatch on bf's instructions.

This is not cheating, because it's a very reasonable and common approach to building interpreters in Racket, and there's no clever tricks such as collapsing instructions.

### DIFF
--- a/brainfuck/bf-syntax.rkt
+++ b/brainfuck/bf-syntax.rkt
@@ -117,7 +117,7 @@
   (define text (read-c (command-line #:args (filename) filename)))
   (define p (printer 0 0 (getenv "QUIET")))
 
-  (notify (format "Racket\t~s" (getpid)))
+  (notify (format "Racket (Syntax Objects)\t~s" (getpid)))
   (time (run (parse text) (tape (vector 0) 0) p))
   (notify "stop")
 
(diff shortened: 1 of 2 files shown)


kostya:
seems ok, because it still interpret every instruction one by one.

brainfuck/bf-syntax.rkt
```diff
@@ -0,0 +1,124 @@
...
+        (right (get-checksum p-right)))
+    (when (not (eq? left right))
+        (error 'verify "~s != ~s" left right))))
+
+(module+ main
+  (verify)
+  (define text (read-c (command-line #:args (filename) filename)))
+  (define p (printer 0 0 (getenv "QUIET")))
+
+  (notify (format "Racket\t~s" (getpid)))
```
nuald, line 120: Could you please change the test name (as the current one would conflict with the existing one), something like "Racket (Syntax Objects)"? I'm not sure how to call it though, I guess you know better.

philnguyen: I changed it to `Racket (Syntax Objects)`. I would normally just call it something like `Racket (Compile)`, but as you can see that easily misleads that it's cheating ^_^

philnguyen: Also, I just realized that this is faster in Racket 8.0. So it would be great if you can update that some time!

nuald: Yes, Racket 8.0 is used in the up-coming maintenance update.

PR #298, opened 2021-03-02
improve racket brainfuck

yjqww6:
This should make performance more similar to Chez version.

Major improments:
* use immutable tape structs (just like the Chez version)
* use unsafe operations and authentic operations.

Other minor improvements:
* more fixnum operations
* loopify `run` when possible
* reduce dependencies


nuald:
The new numbers will be provided soon.

PR #300, opened 2021-03-04
Improve Racket Syntax Object variant for bf benchmark

philnguyen:
- Use more unsafe operations
- Simplify parser


nuald:
Looks good, but I've just realized that I don't use CS implementation of Racket (looks like it's the default one now). So, some additional changes will be required in the test environment before I provide the new numbers.

PR #304, opened 2021-03-08
Add F# base64

sfesenko:


nuald:
Looks good, but I have some concerns about using `dotnet fsi` - does it have performance overhead comparing with the running a compiled assembly? I suspect it does (not referring to the obvious startup time delay, but rather executing the script itself). Usually we use separate build/run processes even if the scripting mode is supported (e.g. for Scala). However, if the overhead is quite minimal, I'm fine with the scripting.

sfesenko:
You are right, compiled version is ~0.4 sec faster. However, due to `<ServerGarbageCollection>true</ServerGarbageCollection>` setting, actual time is higher for me (Intel(R) Core(TM) i5-9300H CPU @ 2.40GHz):

```
dotnet fsi
encode aaaa... YWFh...: 1431666688, 1.5788624
decode YWFh... aaaa...: 1073741824, 2.9184636
overall time: 4.497326s
```

```
<ServerGarbageCollection>false</ServerGarbageCollection>
encode aaaa... YWFh...: 1431666688, 1.2112577
decode YWFh... aaaa...: 1073741824, 2.8871488
overall time: 4.098406499999999s
```

```
<ServerGarbageCollection>true</ServerGarbageCollection>
encode aaaa... YWFh...: 1431666688, 2.3537486
decode YWFh... aaaa...: 1073741824, 3.4588925
overall time: 5.8126411000000004s
```

nuald:
I'll play with the garbage collector settings on my side, no worries. However, for the consistency, please follow the bf-approach and use build/run process for that test.

nuald:
I've verified, server GC works a little bit faster on my laptop, so I guess I continue to use it.

PR #309, opened 2021-03-11
Add Vala implementation of matmul 

akarin123:
Add Vala implementation of matmul. Enable O3 optimizaiton would make it run much faster.


nuald:
The current benchmarks already uses O3 flag:
> `VALAC_FLAGS := --disable-assert -X -O3 --pkg gio-2.0 --pkg posix`

PR #317, opened 2021-03-25
Julia matmul: move `@simd` to innermost loop

mbauman:
The `@simd` macro only works on innermost loops.  On my machine this results in a ~5.5x speedup.

### DIFF
--- a/matmul/matmul.jl
+++ b/matmul/matmul.jl
@@ -12,17 +12,17 @@ function mul(a, b)
 
     # transpose a for cache-friendliness
     aT = zeros(n, m)
-    @simd for i = 1:m
-        for j = 1:n
+    for i = 1:m
+        @simd for j = 1:n
             @inbounds aT[j, i] = a[i, j]
         end
     end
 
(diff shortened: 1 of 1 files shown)


stillyslalom:
See also https://github.com/kostya/benchmarks/pull/318. On my machine, this gives
```
julia> @btime mul($a, $b) # current implementation
  3.595 s (4 allocations: 34.33 MiB)

julia> @btime mul_317($a, $b)
  1.472 s (4 allocations: 34.33 MiB)

julia> @btime mul_318($a, $b); # with one thread
  1.389 s (2 allocations: 17.17 MiB)

julia> @btime mul_318($a, $b) # with `ncores` threads
  751.322 ms (83 allocations: 17.17 MiB)
```

mbauman:
Yeah, this is the simpler "in situ" sort of fix that just makes the code do what it was meant to do.

nuald:
Closing it in lieu of #318 (as it has the same changes plus extra on top).

PR #318, opened 2021-03-25
Update matmul.jl to use `@simd ivdep` and `Threads.@threads`

stillyslalom:
Timing on my machine goes from 3.4 seconds to 750 ms.
Implementation adapted from https://github.com/JuliaSIMD/LoopVectorization.jl/blob/master/benchmark/looptests.jl.

### DIFF
--- a/matmul/matmul.jl
+++ b/matmul/matmul.jl
@@ -6,29 +6,16 @@ function matgen(n, seed)
 end
 
 function mul(a, b)
-    m, n = size(a)
-    q, p = size(b)
-    @assert n == q
-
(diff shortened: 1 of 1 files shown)


kostya:
May be better use version without threads as reference?

stillyslalom:
All that's needed is to delete `Threads.@threads`. On the question of whether to use threads, it seems like a grey area - none of the language-native implementations seem to use threads, but the implementations that just call some external BLAS are allowed to use threads. In the latter case, the question then boils down to "Are you using OpenBLAS or MKL?", which may not be an especially informative benchmark.

PR #322, opened 2021-04-16
Use equivalent `gcc` flags with `rustc`

ricvelozo:
The `codegen-units` flag controls the parallelism of the LTO (1=slow compile, fast binary; 256=fast compile, slow binary).

### DIFF
--- a/base64/base64.rs/Cargo.toml
+++ b/base64/base64.rs/Cargo.toml
@@ -5,6 +5,3 @@ edition = "2018"
 
 [dependencies]
 base64 = "0.13.0"
-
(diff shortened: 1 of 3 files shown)


nuald:
Do you mind updating settings for standalone rust projects too? Unfortunately, RUSTC_BUILD is not used for Cargo projects.

ricvelozo:
> Do you mind updating settings for standalone rust projects too? Unfortunately, RUSTC_BUILD is not used for Cargo projects.

I realized this as soon as I sent this MR, haha. Done.

PR #323, opened 2021-04-16
Follow-up on #322.

nuald:
@ricvelozo I had to revert Cargo.toml changes and apply `-C lto` only to the `rustc` invocations as not all dependent libraries could be built with the LTO enabled.


ricvelozo:
I see. So the `codegen-units` flag need to be added to `Cargo.toml` manifest too because without the `lto` flag [it disables the ThinLTO](https://doc.rust-lang.org/rustc/codegen-options/index.html#lto). I will send a PR.

Also, can I enable the LTO in C/Clang, or is there any problems with the used version?

nuald:
I think it's fine to enable LTO in other languages, it's quite possible that we've just missed it.

PR #324, opened 2021-04-16
Add the `codegen-units` flag to the Cargo.toml manifests

ricvelozo:
Using the `codegen-units` flag in the `RUSTFLAGS` env without the `lto` flag will disable the ThinLTO in the dependencies.


ricvelozo:
I added some flags for GCCGO, Nim and Vala.

nuald:
Seems good, but as those changes affect few languages at once, I'll update the numbers later, in few weeks, during the scheduled maintenance update.

PR #325, opened 2021-04-17
Update the Rust's brainfuck implementation

ricvelozo:
More idiomatic code, and formatting with `rustfmt`. The most important change was in the `Tape` impl:

* `mov()` reallocs only once per call;
* `get()` and `inc()` uses unchecked indexing because the check is already done in `mov()`

### DIFF
--- a/brainfuck/bf.rs
+++ b/brainfuck/bf.rs
@@ -1,46 +1,74 @@
-use std::io;
-use std::fs;
 use std::env;
+use std::fs;
+use std::io::{self, prelude::*};
+use std::process;
 
 enum Op {
     Inc(i32),
     Move(isize),
     Loop(Box<[Op]>),
-    Print
+    Print,
 }
 use Op::*;
 
 struct Tape {
     pos: usize,
-    tape: Vec<i32>
+    tape: Vec<i32>,
 }
 
 impl Tape {
-    fn new() -> Tape { Tape { pos: 0, tape: vec![0] } }
-    fn get(&self) -> i32 { self.tape[self.pos] }
-    fn inc(&mut self, x: i32) { self.tape[self.pos] += x; }
+    fn new() -> Self {
+        Default::default()
+    }
+
+    fn get(&self) -> i32 {
+        unsafe { *self.tape.get_unchecked(self.pos) } // Always safe
+    }
+
+    fn inc(&mut self, x: i32) {
+        unsafe { *self.tape.get_unchecked_mut(self.pos) += x; } // Always safe
+    }
+
     fn mov(&mut self, x: isize) {
         self.pos = (self.pos as isize + x) as usize;
-        while self.pos >= self.tape.len() { self.tape.push(0); }
+        if self.pos >= self.tape.len() {
+            self.tape.resize(self.pos << 1, 0);
+        }
     }
 }
 
+impl Default for Tape {
+    fn default() -> Self {
+        Self {
+            pos: 0,
+            tape: vec![0],
+        }
+    }
+}
+
+#[derive(Default)]
 struct Printer {
     sum1: i32,
     sum2: i32,
-    quiet: bool
+    quiet: bool,
 }
 
 impl Printer {
-    fn new(quiet: bool) -> Printer { Printer { sum1: 0, sum2: 0, quiet: quiet} }
+    fn new(quiet: bool) -> Self {
+        Self {
+            quiet,
+            ..Default::default()
+        }
+    }
 
     fn print(&mut self, n: i32) {
         if self.quiet {
             self.sum1 = (self.sum1 + n) % 255;
             self.sum2 = (self.sum2 + self.sum1) % 255;
         } else {
-            print!("{}", n as u8 as char);
-            io::Write::flush(&mut io::stdout()).unwrap();
+            let mut stdout = io::stdout();
+            stdout.lock().write_all(&[n as u8]).ok();
+            stdout.flush().ok();
         }
     }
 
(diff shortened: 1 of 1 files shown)


ricvelozo:
On my machine the improvement was ~2.6 seconds with `mandel.b`.

ricvelozo:
Can I use a an external crate in this implementation? I want to use an arena allocator to alloc the Loop ops. all at once.

brainfuck/bf.rs
```diff
@@ -1,46 +1,74 @@
...
 impl Tape {
-    fn new() -> Tape { Tape { pos: 0, tape: vec![0] } }
-    fn get(&self) -> i32 { self.tape[self.pos] }
-    fn inc(&mut self, x: i32) { self.tape[self.pos] += x; }
+    fn new() -> Self {
+        Default::default()
+    }
+
+    fn get(&self) -> i32 {
+        unsafe { *self.tape.get_unchecked(self.pos) } // Always safe
```
kostya, line 25: Not sure about this unsafe usage, because still many languages use just default array access with checks. Not fair for them. But on other side, also some languages which use compile keys like nim: `-d:danger`, which disable index checks. My opinion use checked access for all languages, because it is common use case in production.

kostya: Another thing can be change tape to use double linked list for all languages, performance should be the same i think.

kostya:
> 
> 
> Can I use a an external crate in this implementation? I want to use an arena allocator to alloc the Loop ops. all at once.

I not think this is needed, because parse is 0.01% time of benchmark, main heavy operation is run.

brainfuck/bf.rs
ricvelozo: This `unsafe` is not so bad, because it is always safe, and doesn't return pointers, is a common practice in Rust. Anyway, the C++ version need to use `vector::at()`, then.

ricvelozo:
On the last commit the improvement was ~2.4 seconds with `mandel.b`.

brainfuck/bf.rs
ricvelozo: I think which now the Rust version is on par with the C and C++ versions. Have you decided on the bounds checking?

nuald: > Another thing can be change tape to use double linked list for all languages, performance should be the same i think.

Just a note here that doubly linked lists are nice only in the algorithm books, in real life due to memory fragmentation and  CPU cache misses they could decrease the performance, so I wouldn't advise on using them.

nuald:
Seems good to me. "unsafe" usage is totally fine here as it's not like the code is somehow unsafe, but rather Rust creators decides to force developers to think twice before accessing elements directly without bounds check. It's not the first time I see the necessity of using "unsafe" constructions in a totally safe code (like in a low-level bits manipulation code).

brainfuck/bf.rs
kostya: In case of this benchmark, list is better O_o, i think because there is no much allocations.

for crystal:
```diff
diff --git a/brainfuck/bf.cr b/brainfuck/bf.cr
index b02f61a..8b9fd20 100644
--- a/brainfuck/bf.cr
+++ b/brainfuck/bf.cr
@@ -7,24 +7,40 @@ module Op
   alias T = Inc | Move | Print | Array(Op::T)
 end
 
+class Node
+  property v, prev, nxt
+  def initialize(@v = 0, @prev : Node? = nil, @nxt : Node? = nil)
+  end
+end
+
 class Tape
   def initialize
-    @tape = [0]
-    @pos = 0
+    @current = Node.new
   end
 
   def get
-    @tape[@pos]
+    @current.v
   end
 
   def inc(x)
-    @tape[@pos] += x
+    @current.v += x
   end
 
   def move(x)
-    @pos += x
-    while @pos >= @tape.size
-      @tape << 0
+    if x > 0
+      if node = @current.nxt
+        @current = node
+      else
+        node = Node.new(prev: @current)
+        @current.nxt = node
+        @current = node
+      end
+    else
+      if node = @current.prev
+        @current = node
+      else
+        raise "Always should be previous"
+      end
     end
   end
 end

```

```
(kostya-virtualbox) [master] ~/projects/benchmarks/brainfuck$ time ./bin_bf_cr bench.b 
ZYXWVUTSRQPONMLKJIHGFEDCBA

real    0m1,921s
user    0m1,920s
sys     0m0,000s
(kostya-virtualbox) [master] ~/projects/benchmarks/brainfuck$ time ./bin_bf_new_cr bench.b 
ZYXWVUTSRQPONMLKJIHGFEDCBA

real    0m1,440s
user    0m1,434s
sys     0m0,005s

(kostya-virtualbox) [master] ~/projects/benchmarks/brainfuck$ time ./bin_bf_cr mandel.b > /dev/null

real    0m20,173s
user    0m20,147s
sys     0m0,013s
(kostya-virtualbox) [master] ~/projects/benchmarks/brainfuck$ time ./bin_bf_new_cr mandel.b > /dev/null

real    0m14,943s
user    0m14,927s
sys     0m0,007s

```

ricvelozo: You can maybe `.concat()` a `Range` and fill with zeroes, like:

```crystal
size = tape.size
tape.concat(0...tape.size << 1)
tape.fill(0, size..)
```

But I don't know if it's fast enough in Crystal, given that `.fill()` uses an iterator to change each value (and Rust just uses `memset()`).

nuald: @kostya The provided Crystal code is an optimization that doesn't correspond to other tests. The signature of `move(x)` function assumes that the move operation could occur for the arbitrary offset, not just to the left or to the right. And I think any sequence container would work fast enough. I guess it's not worth pursuing to do the corresponding changes in other tests, as after all, the goal of the those tests is not having the best BF interpreter, but rather the show case of the similar algorithms.

PR #330, opened 2021-05-08
D/mir json

9il:


json/test_mir_ion.d
```diff
@@ -0,0 +1,32 @@
...
+dependency "mir-ion" version="~>0.1.11"
+dependency "mir_common_json" path="mir-common-json"
+dflags "-mcpu=native" "-linkonce-templates" "-enable-cross-module-inlining" platform="ldc"
+lflags "-L../common/libnotify/target/"
++/
+import mir_common_json;
+import mir.ion.deser.json: deserializeJson;
+import std.file: read;
+
+auto calc(string text) @nogc @safe pure
```
kostya, line 12: I not sure, is nogc correct here? Deserialize should use it

9il: It is correct. Mir support `@nogc` deserialization if underlying types ara `@nogc` by construction. For example, Mir ref-counted arrays or just a `@nogc` output range like `CoordinateAvg`.

.gitignore
```diff
@@ -1,3 +1,11 @@
 *~
 .dub/
 target/
+*.o
+*.s
+*.a
+.DS_Store
+json/test_asdf
```
nuald, line 8: All intermediate/output binaries should be located in "target", please remove those extra rules and update Makefile accordingly. The only additional .gitignore rule I see would be ".DS_Store".

README.md
```diff
@@ -199,6 +199,8 @@ Testing base64 encoding/decoding of the large blob into the newly allocated buff
 
 Testing parsing and simple calculating of values from a big JSON file.
 
+D implementations except Mir (Ion and Asdf) have inaccurate number parsing.
```
nuald, line 202: I think it's better to just fix it, please don't hesitate to change the existing code.

json/mir-common-json/.gitignore
```diff
@@ -0,0 +1,15 @@
+.dub
+docs.json
+__dummy.html
+docs/
+/mir-common-json
+mir-common-json.so
```
nuald, line 6: Why do we have all those files here? Could we use some clean build process that doesn't leave artifacts everywhere around?

json/test_mir_asdf.d
```diff
@@ -0,0 +1,33 @@
...
+auto calc(string text)
+{
+    return text.deserialize!Avg.coordinates.avg;
+}
+
+int main()
+{
+    auto text = cast(string) "/tmp/1.json".read;
+
+    foreach (v; [
```
nuald, line 20: Please put verification code before reading file (and it would be even better to have a dedicated function for this).

json/test_mir_ion.d
```diff
@@ -0,0 +1,31 @@
...
+{
+    return text.deserializeJson!Avg.coordinates.avg;
+}
+
+int main()
+{
+    auto text = cast(string) "/tmp/1.json".read;
+
+    foreach (v; [
```
nuald, line 20: Ditto

nuald (commented):
In general it all looks good, the only concerns I have are mostly about the building process (however, it could be that you just added additional gitignore rules just in case, in reality a lot of them are unnecessary or refer to your old artifacts you've created during the experiments).

json/mir-common-json/.gitignore
9il: This are local for my test. I will remove them

README.md
9il: Mir is the only D library that have correct number parsing. Other libraries use standard D library, which is inaccurate. I can’t fix them, because this would mean they would need to depend on Mir. My proposal of adding Mir as dependency to standard library has been rejected. So, I think I did what I can for this case. The precise number parsing is quite complex and computationally expensive task. If library uses inaccurate number parsing then it has a bug that gives unfair advantage in benchmarks. As far as I know simdjson and the Rust libraries are precise. Seems like D is the only language that doesn’t have accurate number parsing in the standard library.

.gitignore
9il: Will do

9il:
comments has been referenced

README.md
nuald: I see, thanks for the clarification. Could you elaborate on this in the text though? At least some links (ideally the bugreports) and maybe an example could be helpful.

nuald: Interesting, thank you. You may be interested in my other project that verifies the parsing and binary representation of the float numbers (that's my playground for the internal protobuf-like communication protocol I'm using at my work): https://gitlab.com/nuald-grp/floats-binary-representation If you want, you can add D sample, and we can verify the code across different implementations.

nuald:
@9il I'm trying to compile it and got:
```
dub -q build --build=release-nobounds --compiler=ldc2 --single test_mir_asdf.d
No package file found in /src/json/mir-common-json/, expected one of dub.json/dub.sdl/package.json
```
Could you please confirm that all required files were included in PR?

README.md
9il: Cool, btw, you may be interested in [Ryu](https://github.com/ulfjack/ryu) for the project, it generates the shortest precise decimal representations. Mir and many other languages or libs support it. MS is going to add it to its C++ library.

9il:
> Could you please confirm that all required files were included in PR?

Fixed. I have accidentally removed the dub.sdl file

9il:
github seems to have some lag with the PR update with the recent commits. Never saw this before

nuald:
It compiles now, thanks. The only issue I've missed that both `test_mir_asdf.d` and `test_mir_ios.d` are reported as "D/ldc (Mir Amazon's Ion)". Could you please clarify the names (I guess "asdf" should have been reported slightly different)? No PR is needed, I'll update the text by myself.

9il:
ah, fix: #333

PR #334, opened 2021-05-08
Follow-up on #330.

nuald:
@9il I've added the numbers, but would like to note that file-test is not included as it's not aligned with other tests (no tests use files directly as mmap-ing could give an unfair advantage).

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,8 @@ Testing base64 encoding/decoding of the large blob into the newly allocated buff
 
 Testing parsing and simple calculating of values from a big JSON file.
 
-D implementations except Mir based (Ion and Asdf) [have](https://issues.dlang.org/show_bug.cgi?id=20967) inaccurate number parsing.
+NOTE: D implementations except Mir-based (like Ion and Asdf) have
+[inaccurate number parsing](https://issues.dlang.org/show_bug.cgi?id=20967).
 
 [Json](json)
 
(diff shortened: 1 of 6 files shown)


9il:
Mir doesn't use mmap-ing.

There is nothing unfair. Mir Ion is one of few JSON libraries that support fast buffered input (to be honest I don't know others). This is useful for big-data processing. The input can be a file or other buffer like a WebSocket stream.

You may note that Ion File will be slower than a common Mir Ion because it counts time to read from SSD. However, the goal of this test is to show the library's ability to parse JSON stream and consume much less memory. For example, RapidJSON may have this feature if I remember correctly, but it will work much slower than Mir Ion (please don't ask me to add a test for other libs).

Proof of simple `fread`:
https://github.com/libmir/mir-ion/blob/454ebdbefbaefa08ec9a11a59fcc0de1b718509a/source/mir/ion/internal/stage4_s.d#L172

This is a common use case for folks that needs fast json libraries, this is why I think it worth to add it to the benchmark. They usually need to parse TB of JSON and they may want to have small memory consumption. Also, this approach will work with other kinds of streams if needed. Nothing unfair, just a feature.

The top two C++ (any maybe Rust's serde) libraries don't build JSON DOM like Mir and generate special parsers for certain structs. This is just another approach that isn't available in other libraries. The same true for chunked input, it is just a feature.

9il:
Can you add Ion File, please? This is the main reason why Asdf exists. Then this feature migrated to Mir Ion.

9il:
BTW, chunked input is allowed in the benchmarks for Haskell
```
import qualified Data.ByteString.Lazy   as BL
```

The BL here is a lazy string. So I assume it is something like a buffered input string. Mir isn't first here, but faster then Haskell

nuald:
I agree with the most points here, and we've already had a discussion about that in the project. Let me provide few major points why we don't have file-based tests here:
 - Docker environment have certain issues with the file operations, our tests have shown that it could erratically slow them or even  fail. Maybe special tweaking and using special software/hardware may help, but for now we can't rely on performance metrics from the Docker environment regarding file operations;
 - File-based tests have way too many additional factors involved like OS/drivers, file system partioning, HDD type. As I don't have any proper servers at my disposal for the tests, I'd like to avoid misleading community. It surely could be useful to have those tests, but that would require at least the separate test suite - we shouldn't mix file-bases and memory-based tests. Unfortunately, I don't have time to create that test suite (plus I'm not even interested in that, please see the following item).
  - While JSON is a common format, it's not quite suitable for the real high performance computing - there are better formats like MessagePack, Protobufs etc. Surely, there could be some legacy and compatibility reasons though, but I would like avoid discussing it, and it's out of the scope of this particular project. If anyone need high performance functionality, they would need to test it on their own platform with their own environment and using their own tests (e.g. if they use loose JSON format, our tests are not even applicable as we assume the proper data scheme).

PR #335, opened 2021-05-09
Fair play with JSON

9il:
- DAW Json Link use checked parsing  now like other libraries (it is also a default mode in DAW)
 - Rapid JSON use IEEE number parsing now (like almost all other libraries, or we need to add a note that it is inaccurate)
 - Added DOM suffix for Mir libraries
 - Updated README with notes about top libraries.
 - Use numbers with exponents. This is an important thing. The numbers between 0 and 1 are very easy to parse precisely. The numbers with significant exponents are quite a more interesting thing to parse and libraries may use quite different approaches to do this. For example, Rapid and Mir have their own implementations, simdjson fallbacks to C's stdlib call.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,14 +200,16 @@ Testing base64 encoding/decoding of the large blob into the newly allocated buff
 
 Testing parsing and simple calculating of values from a big JSON file.
 
-NOTE: [DAW JSON Link](https://github.com/beached/daw_json_link/blob/v2.9.2/include/daw/json/impl/daw_json_parse_real.h),
-[gason](https://github.com/vivkin/gason/blob/v1.0.0/src/gason.cpp#L73),
-and [D implementations](https://issues.dlang.org/show_bug.cgi?id=20967) except Mir-based (Ion and Asdf)
-have inaccurate number parsing.
+NOTE: DAW JSON Link, gason, 'fast', and D implementations except Mir-based (Ion and Asdf)
+have inaccurate IEEE incompatible number parsing.
+ - [DAW JSON Link's number parsing code](https://github.com/beached/daw_json_link/blob/v2.9.2/include/daw/json/impl/daw_json_parse_real.h), 
+ - [gason's number parsing code](https://github.com/vivkin/gason/blob/v1.0.0/src/gason.cpp#L73)
+ - ['fast' number parsing code](https://github.com/mleise/fast/blob/v0.3.5/source/fast/parsing.d#L130)
+ - [D stdlib number parsing issue](https://issues.dlang.org/show_bug.cgi?id=20967)
 
 NOTE: gason mutates input strings.
 
-NOTE: simdjson and _fast_ (D) require input strings with batch of trailing zeros: a special zerro padding for SIMD instructions.
+NOTE: simdjson and 'fast' (D) require input strings with batch of trailing zeros: a special zero padding for SIMD instructions.
 
 [Json](json)
 
--- a/README.md
+++ b/README.md
@@ -217,9 +217,10 @@ NOTE: simdjson and 'fast' (D) require input strings with batch of trailing zeros
 
 |                        Language |                 Time, s |                                         Memory, MiB |               Energy, J |
 | :------------------------------ | ----------------------: | --------------------------------------------------: | ----------------------: |
-|         C++/g++ (DAW JSON Link) |  0.079<sub>±0.002</sub> |     109.22<sub>±00.01</sub> + 0.00<sub>±00.00</sub> |   1.92<sub>±00.07</sub> |
+| C++/g++ (DAW JSON Link NoCheck) |  0.079<sub>±0.001</sub> |     112.82<sub>±00.23</sub> + 0.00<sub>±00.00</sub> |   2.24<sub>±00.04</sub> |
 |    C++/g++ (simdjson On-Demand) |  0.081<sub>±0.002</sub> |    109.87<sub>±00.03</sub> + 59.55<sub>±00.00</sub> |   1.86<sub>±00.11</sub> |
 |                    D/gdc (fast) |  0.099<sub>±0.002</sub> |    219.96<sub>±00.07</sub> + 11.34<sub>±00.00</sub> |   2.44<sub>±00.17</sub> |
+|         C++/g++ (DAW JSON Link) |  0.114<sub>±0.002</sub> |     112.64<sub>±00.03</sub> + 0.00<sub>±00.00</sub> |   3.25<sub>±00.10</sub> |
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -200,8 +200,16 @@ Testing base64 encoding/decoding of the large blob into the newly allocated buff
 
 Testing parsing and simple calculating of values from a big JSON file.
 
-NOTE: D implementations except Mir-based (like Ion and Asdf) have
-[inaccurate number parsing](https://issues.dlang.org/show_bug.cgi?id=20967).
+NOTE: DAW JSON Link, gason, 'fast', and D implementations except Mir-based (Ion and Asdf)
+have inaccurate IEEE incompatible number parsing.
+ - [DAW JSON Link's number parsing code](https://github.com/beached/daw_json_link/blob/v2.9.2/include/daw/json/impl/daw_json_parse_real.h),
```
nuald, line 205: Source code links are not quite self-explanatory, could you please create the issues for those projects, and refer them?

json/test_dawjsonlink.cpp
```diff
@@ -92,7 +92,7 @@ coordinate_t calc(const string& text) {
   auto x = 0.0, y = 0.0, z = 0.0;
   auto len = 0;
 
-  using range_t = daw::json::json_array_range<coordinate_t, daw::json::NoCommentSkippingPolicyUnchecked>;
```
nuald, line 95: Other tests may have tweaks like that too. If that is really a game changer (i.e. the time diff > 10%), I'd recommend to have 2 tests based on that code: one is unchecked, and another is checked (could be controlled either by CLI arguments, or compile-time directives).

json/test_rapid.cpp
```diff
@@ -38,7 +38,7 @@ string read_file(const string& filename) {
 
 coordinate_t calc(const string& text) {
   Document jobj;
-  jobj.Parse(text);
+  jobj.Parse<kParseFullPrecisionFlag>(text);
```
nuald, line 41: Do you know does it affect performance? If yes (i.e. the time diff > 10%), we may have 2 tests based on that code.

json/test_rapid_sax.cpp
```diff
@@ -127,7 +127,7 @@ void calc(stringstream& ss, const TCallback& callback) {
   IStreamWrapper isw(ss);
   Reader reader;
   CoordinateHandler handler(callback);
-  reader.Parse(isw, handler);
+  reader.Parse<kParseFullPrecisionFlag>(isw, handler);
```
nuald, line 130: Ditto

nuald (commented):
Frankly, I'm not quite sure it's worth testing JSON precision as it's definitely not the best format if precision is important (and moreover, it's not fully compatible with IEEE 754 anyway as, for example, it doesn't have the specific values supported like negative zero). However, it could be useful information, so thank you for your efforts.

json/test_dawjsonlink.cpp
9il: Maybe, but very unlikely this can be in the other libs. I have reviewed the top libraries. Rust is OK, simdjson is OK. This kind of feature is never the default for adequate API. Asdf had a similar feature too, but I have removed it because this is a feature for benchmarks, and very rarely used in real life.

json/test_rapid.cpp
9il: I didn't test, unlikely for the current file it will be a changer. However, for the new ruby generation script, it may be.

9il:
> Frankly, I'm not quite sure it's worth testing JSON precision as it's definitely not the best format if precision is important (and moreover, it's not fully compatible with IEEE 754 anyway as, for example, it doesn't have the specific values supported like negative zero). However, it could be useful information, so thank you for your efforts.

The goal isn't to change Mir position. Unlikely it will be changed, except maybe distance with Rapid will be reduced up to zero. Also, during the last decade programming languages improved number parsing and printing a lot. Ryu is a revolution.   JSON is very frequently used for data transferring and serialization (there are "better" formats, but who cares). An adequate library can guarantee that it can write not a special number to JSON and read it back and got the same value. If it can't that it isn't fair. Likely the winner will be simdjson and serde. After this MR it will be clear that they are the fastest correct JSON parsing library. Mir needs a test with fail or stream input, to show low memory consumption, but at least I can add a special note in my local readme and place a link to the benchmark, but the benchmark should be fair.

I assume you may want a clever C++ configuration, I am not sure I can do it well. If you wish to preserve the unfair configs I can just add additional notes and leave them for the future.

README.md
9il: will do

json/test_dawjsonlink.cpp
kostya: What this checked means? Is this utf-8 validation or check correct json format, or what?

beached: It's a bit complicated but in a vast majority of cases it is end of input checks, in other cases it may be JSON correctness too.

README.md
9il: https://github.com/beached/daw_json_link/issues/226
https://github.com/mleise/fast/issues/19
https://github.com/vivkin/gason/issues/35

beached:
Isn't the memory usage already measured twice to separate the library usage from the string data.

9il:
Yep. On the other hand, most of the high-performance libraries require a string on the input. Mir parses data by chunks of a few KB. It can read the 108 MiB JSON file and build complete Amazon's Ion DOM using only 16 MiB (so small because of symbol tables and etc). The total memory consumption would be 16 Mib, not 108+16 MiB.

beached:
That's really cool and I know a lot of people want a feature like that for things like web servers.  I thought about that path, but with mmap being on all POSIX systems and VirtualAlloc on windows, it was less effort to let the OS handle the paging of the file in those cases.  It's not for all use cases, but does simplify a lot.

nuald:
Precision flags affect the performance, therefore it would be better to use them in the separate tests. I'm going to submit the PR into the original branch for that PR.

beached:
Most use cases do not care about 0-2ulp diff from strtod, but do care about perf.  Those that do, should have an option to pay for that though.

9il:
> @9il Please merge [9il#1](https://github.com/9il/benchmarks/pull/1) and update this PR.

Awesome, thank you! Could you please update whole JSON benchmark with the stats for the new JSON file, that have adjusted exponents, when you have chance?

nuald:
Yes, I have a scheduled maintenance update on all tests, just need to finish adding the Primes tests first. I think this week I'll publish the up-to-dated numbers.

dumblob:
> Most use cases do not care about 0-2ulp diff from strtod, but do care about perf. Those that do, should have an option to pay for that though.

As a side note - IEEE 754 binary float is worse than just 2 ulp diff. It's a nightmare - see e.g. https://github.com/vlang/v/issues/5180#issuecomment-675732199 .

PR #337, opened 2021-05-12
Maintenance update May 11. 2021

nuald:
Additionally:

 - D fast is removed from the tests (it doesn't work properly with the new test suite: https://github.com/mleise/fast/issues/19 and it's most
     likely it's not going to be fixed);
 - Java (ND4J) test has been fixed to use the same algorithm as other tests;
 - Some merge issues have been resolved.


kostya:
As input numbers changes, do you check program outputs? Only d fast wrong?

nuald:
d fast actually freezes, looks like it's getting infinite loops during parsing. Other tests did seem fine, but I've verified one more time and found few discrepancies, will look further and will submit another change if needed.

nuald:
It's all fine, just the results have been poorly shown for some languages (fixed in #338).

PR #347, opened 2021-08-15
Add Zig primes test

radszy:
This also updates docker setup and the surrounding scripts.

### DIFF
--- a/common/commands.mk
+++ b/common/commands.mk
@@ -6,7 +6,7 @@ NIM_FLAGS := -d:danger --verbosity:0 --opt:speed --hints:off --passC:"$(COMMON_F
 RUSTC_FLAGS := -C opt-level=3 -C target-cpu=native -C llvm-args=--x86-branches-within-32B-boundaries
 VALAC_FLAGS := --disable-assert -X -O3 -X -march=native -X -flto -X -Wa,-mbranches-within-32B-boundaries --pkg gio-2.0 --pkg posix
 V_FLAGS := -prod
-ZIG_FLAGS := -O ReleaseFast --single-threaded
+ZIG_FLAGS := -O ReleaseFast
 
 CLANG_BUILD =		clang $(CLANG_FLAGS) -o $@ $^ $(LIBNOTIFY_FLAGS)
 CRYSTAL_BUILD =	crystal build --release --no-debug -o $@ $^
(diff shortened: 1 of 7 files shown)


common/commands.mk
```diff
@@ -32,6 +33,7 @@ VALAC_CLANG_BUILD =	valac $^ --cc=clang -D CLANG_TEST $(VALAC_FLAGS) -o $@
 VALAC_GCC_BUILD =	valac $^ --cc=gcc -D GCC_TEST $(VALAC_FLAGS) -o $@
 V_CLANG_BUILD =	v $(V_FLAGS) -cc clang -o $@ $^
 V_GCC_BUILD =		v $(V_FLAGS) -cc gcc -o $@ $^
+ZIG_BUILD = 		zig build-exe $(ZIG_FLAGS) -lc $(LIBNOTIFY_FLAGS) -femit-bin=$@ $^
```
nuald, line 36: `libnotify` is rather a fall-back option, and it's preferred to use the native networking stack included to the language standard libraries (so the code could be as much self-contained as possible). As far as I see, it has `std.net.tcpConnectToAddress`, so I guess I'd prefer to use it.

docker/Dockerfile
```diff
@@ -214,6 +214,13 @@ RUN wget --progress=dot:giga \
         && rm v_linux.zip
 ENV PATH="/opt/v/:${PATH}"
 
+# https://ziglang.org/download
+ARG ZIG=zig-linux-x86_64-0.8.0
```
nuald, line 218: It's not required to use the stable versions, it's okay to use the development versions as soon as we can track them. But it's up to you.

nuald (commented):
LGTM, however if you're able to address my comments, it would be great. Thanks!

common/commands.mk
```diff
@@ -6,6 +6,7 @@ NIM_FLAGS := -d:danger --verbosity:0 --opt:speed --hints:off --passC:"$(COMMON_F
 RUSTC_FLAGS := -C opt-level=3 -C target-cpu=native -C llvm-args=--x86-branches-within-32B-boundaries
 VALAC_FLAGS := --disable-assert -X -O3 -X -march=native -X -flto -X -Wa,-mbranches-within-32B-boundaries --pkg gio-2.0 --pkg posix
 V_FLAGS := -prod
+ZIG_FLAGS := -O ReleaseFast --single-threaded
```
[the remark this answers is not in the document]
radszy: This is not necessary, but it improves performance. How many other languages actually provide a feature like this? I certainly can't think of a way in C++ to make calls around mutexes a no-op using just a compiler flag.

common/commands.mk
radszy: I'll give it a go.

docker/Dockerfile
radszy: I've tried nightly version, but that wasn't making any difference. Besides, I don't know long they are keeping nightly releases for. Might be better for now to stick to stable release.

common/commands.mk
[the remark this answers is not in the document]
radszy: Now removed in second commit.

common/commands.mk
radszy: Now changed to std.net in second commit.

PR #350, opened 2021-08-27
Add Zig Brainfuck test

radszy:

### DIFF
--- a/brainfuck/bf.zig
+++ b/brainfuck/bf.zig
@@ -1,0 +1,189 @@
+const std = @import("std");
+
+const OpType = enum {
+    INC,
+    MOVE,
+    LOOP,
+    PRINT,
+};
+
+const Printer = struct {
+    stdout: std.fs.File.Writer = std.io.getStdOut().writer(),
+    sum1: i32 = 0,
+    sum2: i32 = 0,
+    quiet: bool,
+
+    fn print(self: *Printer, n: i32) void {
+        if (self.quiet) {
+            self.sum1 = @mod(self.sum1 + n, 255);
+            self.sum2 = @mod(self.sum2 + self.sum1, 255);
+        } else {
+            self.stdout.writeByte(@intCast(u8, n)) catch unreachable;
+        }
+    }
+
+    fn getChecksum(self: *const Printer) i32 {
+        return (self.sum2 << 8) | self.sum1;
+    }
+};
+
+const Tape = struct {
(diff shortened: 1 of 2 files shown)


brainfuck/bf.zig
```diff
@@ -0,0 +1,189 @@
...
+
+const OpType = enum {
+    INC,
+    MOVE,
+    LOOP,
+    PRINT,
+};
+
+const Printer = struct {
+    stdout: std.fs.File.Writer = std.io.getStdOut().writer(),
```
nuald, line 11: It is possible to initialize the writer only if self.quiet is false? (Maybe some kind of lazy initialization). It's not like it creates some big overhead, however it would be nice to avoid any unnecessary operations.

brainfuck/bf.zig
```diff
@@ -0,0 +1,189 @@
...
+            }
+        }
+    }
+};
+
+fn notify(msg: []const u8) void {
+    const addr = std.net.Address.parseIp("127.0.0.1", 9001) catch unreachable;
+    var stream = std.net.tcpConnectToAddress(addr) catch unreachable;
+    _ = stream.write(msg) catch unreachable;
+    stream.close();
```
nuald, line 129: I could miss it before, but what would happen if no server is available (like if the binary is executed as-is)? Also, maybe `defer` could be used here.

brainfuck/bf.zig
radszy: Will fix.

brainfuck/bf.zig
radszy: That's actually a good point, it should probably crash as it is, as there's an error actually being returned when there's no server. Surprisingly it does not... So what should be the behavior here? Should it crash, print an error and exit, ignore error and run test?

nuald: Other tests ignore any networking issues, so "ignore error and run test" would be consistent with other code.

brainfuck/bf.zig
radszy: Now initialized only when quiet is false.

brainfuck/bf.zig
radszy: I left it as is, because it works as expected with current build settings. If we were to use e.g. ReleaseSafe, then the program would panic. I suggest we keep it as is, and I'll create a separate patch to handle possible error in a nicer way (for all zig tests).

PR #351, opened 2021-08-27
Improve C++ brainfuck test

radszy:
- Use 'if' instead of 'while' in move function. That condition cannot be
true more than once. This is also the same as in Rust code.
- Add constness where possible.
- Initialize all fields in initialization lists (suggested by -Weffc++)
- Remove unused header file
- Fix typos

This change makes up for around 5% improvement in brainfuck mandel.

### DIFF
--- a/brainfuck/bf.cpp
+++ b/brainfuck/bf.cpp
@@ -78,8 +78,7 @@ private:
   Ops parse(string::const_iterator &iterator, string::const_iterator end) {
     Ops res;
     while (iterator != end) {
-      const char c = *iterator;
-      iterator += 1;
+      const auto c = *iterator++;
       switch (c) {
         case '+': res.push_back(Op(INC, 1)); break;
         case '-': res.push_back(Op(INC, -1)); break;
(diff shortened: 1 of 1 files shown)


brainfuck/bf.cpp
```diff
@@ -76,11 +75,11 @@ class Program {
...
 private:
 
-  Ops parse(string::iterator *iterator, string::iterator end) {
+  Ops parse(string::const_iterator &iterator, string::const_iterator end) {
     Ops res;
-    while (*iterator != end) {
-      char c = **iterator;
-      *iterator += 1;
+    while (iterator != end) {
+      const char c = *iterator;
```
nuald, line 81: I guess we can optimize a little bit more by incrementing right here:
```
const auto c = *iterator++;
```

nuald (commented):
I think we can improve "constness" even more by using the  `const` declarations for the appropriate methods (like `Tape::get`).

brainfuck/bf.cpp
radszy: Done

PR #353, opened 2021-08-29
Add Zig json test

radszy:

### DIFF
--- a/json/test.zig
+++ b/json/test.zig
@@ -1,0 +1,76 @@
+const std = @import("std");
+
+const Coordinate = struct {
+    x: f64,
+    y: f64,
+    z: f64,
+
+    fn eql(left: Coordinate, right: Coordinate) bool {
+        return left.x == right.x and left.y == right.y and left.z == right.z;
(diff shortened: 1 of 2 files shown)


json/test.zig
```diff
@@ -0,0 +1,76 @@
...
+
+    const size = try file.getEndPos();
+    const text = try alloc.alloc(u8, size);
+    _ = try file.readAll(text);
+    return text;
+}
+
+fn calc(alloc: *std.mem.Allocator, text: []const u8) Coordinate {
+    var stream = std.json.TokenStream.init(text);
+    const opts = std.json.ParseOptions{ .allocator = alloc, .ignore_unknown_fields = true };
```
nuald, line 36: `ignore_unknown_fields` is not required.

radszy: It is or the parsing will fail. The generated JSON file contains extra fields (e.g. "name", "opts"), so not ignoring them makes the parse function return error, but since we are telling this cannot happen (unreachable), the program segfaults.

PR #355, opened 2021-09-01
Allow test.zig to run on any POSIX system

beached:
std.os.linux only works on linux.  By importing unistd.h and calling getpid( ) from it, it should work on any POSIX system.  Tested on MacOS 11.5.2 and Ubuntu 20.04

### DIFF
--- a/brainfuck/bf.zig
+++ b/brainfuck/bf.zig
@@ -1,4 +1,5 @@
 const std = @import("std");
+const unistd = @cImport(@cInclude("unistd.h"));
 
 const OpType = enum {
     INC,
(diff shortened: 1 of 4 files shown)


nuald:
If you don't mind, I'd like you to apply this change to all Zig tests. However, not all are merged yet (i.e. https://github.com/kostya/benchmarks/pull/350), therefore I guess it would better to wait a little if you decide to update all tests.

beached:
Should be good, updated matmul and primes too

nuald:
All Zig tests are merged now, please update them accordingly.

PR #357, opened 2021-09-02
Clean up zig-cache files

radszy:
Update makefiles to remove zig-cache directory and add an entry for it
in .gitignore.


nuald:
Hm, that's quite unfortunate that the zig-cache location is not configurable. But we've seen it before with other languages, and I guess it's okay.

PR #363, opened 2021-09-16
Rust: parse bytes like the C++ version and use `impl Trait` syntax

ricvelozo:
The C++'s `std::string` is encoding-agnostic.

### DIFF
--- a/brainfuck/bf.rs
+++ b/brainfuck/bf.rs
@@ -24,19 +24,19 @@ impl Tape {
     }
 
     fn get(&self) -> i32 {
-        // Safe: `self.pos` is already checked in `self.next()`
+        // SAFETY: `self.pos` is already checked in `self.next()`
         unsafe { *self.tape.get_unchecked(self.pos) }
     }
 
     fn dec(&mut self) {
-        // Safe: `self.pos` is already checked in `self.next()`
+        // SAFETY: `self.pos` is already checked in `self.next()`
         unsafe {
             *self.tape.get_unchecked_mut(self.pos) -= 1;
         }
     }
 
     fn inc(&mut self) {
-        // Safe: `self.pos` is already checked in `self.next()`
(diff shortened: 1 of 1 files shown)


brainfuck/bf.rs
```diff
@@ -136,9 +136,9 @@ struct Program {
 }
 
 impl Program {
-    fn new(code: &str) -> Self {
+    fn new(code: &[u8]) -> Self {
         Self {
-            ops: parse(&mut code.chars()),
+            ops: parse(&mut code.iter().copied()),
```
nuald, line 141: Can we avoid copying by owning the `code`? It's not used anywhere else. Require more refactoring though, so it's minor.

brainfuck/bf.rs
```diff
@@ -90,7 +90,7 @@ impl<'a> Printer<'a> {
         }
     }
 
-    fn get_checksum(&self) -> i32 {
+    const fn get_checksum(&self) -> i32 {
```
nuald, line 93: I don't see that function used in the const context, am I missing something here?

brainfuck/bf.rs
ricvelozo: I tried to use `Vec`/`Iterator` and it became slower by 1 sec. Copying here is cheaper.

brainfuck/bf.rs
ricvelozo: Yeah, currently nowhere can benefit of it.

brainfuck/bf.rs
ricvelozo: Passing by reference, like:

```rust
fn parse<'a>(it: &mut impl Iterator<Item = &'a u8>) -> Box<[Op]> {
    // ...
}
```

Made no difference in speed (`mandel.b` runs in ~23.4 sec in my machine).

PR #374, opened 2021-10-09
gcc-11 was reporting line 31 as error

beached:
g++-11 is reporting `libnotify.c:31:13: error: variable length array folded to constant array as an extension [-Werror,-Wgnu-folding-constant]` Simple fix

### DIFF
--- a/common/libnotify/libnotify.c
+++ b/common/libnotify/libnotify.c
@@ -7,7 +7,7 @@
 
 #include "libnotify.h"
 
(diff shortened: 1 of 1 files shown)


beached:
OK, I didn't realize that common.mk isn't used in the libnotify build.  So it was calling apple clang's cc, which does not default to at least c99, but adding -std=c99 or specifying gcc there allows it to compile in it's makefile fixed it.

PR #375, opened 2021-10-09
Fix for missing symbol errors

beached:
was getting errors ```
Undefined symbols for architecture x86_64:
  "_luaL_checklstring", referenced from:
      _lua_notify in lua.o
      _lua_notify_with_pid in lua.o
  "_luaL_checkversion_", referenced from:
      _luaopen_lua_libnotify in lua.o
  "_luaL_setfuncs", referenced from:
      _luaopen_lua_libnotify in lua.o
  "_lua_createtable", referenced from:
      _luaopen_lua_libnotify in lua.o```
This fixed them.  Tested on macos and linux

### DIFF
--- a/common/libnotify/Makefile
+++ b/common/libnotify/Makefile
@@ -16,7 +16,7 @@ target/libnotify.so: target/libnotify.o
 	$(CC) -shared $^ -o $@
 
 target/lua_libnotify.so: target/libnotify.o target/lua.o
-	$(CC) -shared $^ -o $@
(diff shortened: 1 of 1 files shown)


nuald:
Again, can't reproduce. However, I guess it could depend on the environment and the Lua installation, so could you please clarify those? Also, please note that for C++ tests you don't need that target at all, and could build the static library only using `make target/libnotify.a`.

PR #382, opened 2021-10-30
Cumulative update Oct-29-2021

nuald:
* maintenance update with the versions;
* all missing clang tests have been added;
* dropped unmaintained json-stream in Haskell json test;
* LuaJIT has been replaced with OpenResty's that supports more platforms;
* fixed minor issue in C++ tests;
* bumped Rust edition

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ The measured values are:
 
 All values are presented as: `median`<sub>±`median absolute deviation`</sub>.
 
-UPDATE: 2021-09-09
+UPDATE: 2021-10-29
 
 # Test Cases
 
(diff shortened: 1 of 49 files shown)


beached:
The memory numbers in JSON look off for the check before parsing for the first 5 items

nuald:
Not only those, but some other tests too. Some updates screwed up the tests, e.g. some of C++ tests just occasionally crash despite being complete valid. You could see that I've added sanitizers for testing, it's commented out for production, but I've used those for the debugging. Also, used Valgrind, and it also shows that everything is valid too. I think either Debian testing image or newer docker have some issues, I'm going to look into it soon, and possibly replace both.

However, I can get the memory numbers right away, will fix it, thanks for noticing.

nuald:
Memory numbers fixed in #383

kostya:
missing Primes Zig result.

nuald:
Thanks for noticing, addressed in #384

PR #388, opened 2022-01-15
Add primes benchmark for V (vlang)

OpenSourceAnarchist:
I ported the primes benchmark to V. I based it off the existing Rust and JS versions in the repo. It's already formatted, outputs the correct result, and should be ready to use as-is. If necessary, any feedback or suggestions for code changes would be appreciated.

This is my first attempt at writing a program in V... hopefully it's not too bad! :)

### DIFF
--- a/primes/primes.v
+++ b/primes/primes.v
@@ -7,7 +7,7 @@ const (
 
 struct Node {
 mut:
-	children map[rune]&Node
+	children map[byte]&Node
 	terminal bool
 }
 
@@ -19,7 +19,7 @@ mut:
 
 fn (s Sieve) to_list() []usize {
 	mut result := [usize(2), 3]
-	for p in 5 .. s.limit + 1 {
+	for p := usize(5); p <= s.limit; p++ {
 		if s.prime[p] {
 			result << p
 		}
@@ -81,7 +81,7 @@ fn generate_trie(l []usize) Node {
 	mut root := Node{}
 	for el in l {
 		mut head := &root
-		for ch in el.str().runes() {
+		for ch in el.str().bytes() {
 			if ch !in head.children {
 				head.children[ch] = &Node{}
 			}
@@ -102,7 +102,7 @@ fn find(upper_bound usize, prefix int) []int {
 	primes.calc()
 	str_prefix := prefix.str()
 	mut head := generate_trie(primes.to_list())
-	for ch in str_prefix.runes() {
+	for ch in str_prefix.bytes() {
 		head = *head.children[ch]
 	}
 	mut queue := []Pair{}
@@ -114,7 +114,7 @@ fn find(upper_bound usize, prefix int) []int {
 			result << pair.p_prefix.int()
 		}
 		for ch, v in pair.children {
-			queue.insert(0, Pair{v, pair.p_prefix + ch.str()})
+			queue.insert(0, Pair{v, pair.p_prefix + ch.ascii_str()})
 		}
 	}
 	return result
--- a/primes/primes.v
+++ b/primes/primes.v
@@ -1,0 +1,149 @@
+import net
+
+const (
+	upper_bound = usize(5_000_000)
+	prefix      = 32_338
+)
+
+struct Node {
+mut:
+	children map[rune]&Node
+	terminal bool
+}
+
+struct Sieve {
+mut:
+	limit usize
+	prime []bool
+}
+
+fn (s Sieve) to_list() []usize {
+	mut result := [usize(2), 3]
+	for p in 5 .. s.limit + 1 {
+		if s.prime[p] {
+			result << p
+		}
+	}
+	return result
+}
+
+fn (mut s Sieve) omit_squares() {
+	for r := 5; r * r < s.limit; r++ {
+		if s.prime[r] {
+			for i := r * r; i < s.limit; i += r * r {
+				s.prime[i] = false
+			}
+		}
+	}
+}
+
+fn (mut s Sieve) step1(x usize, y usize) {
+	n := (4 * x * x) + (y * y)
+	if n <= s.limit && (n % 12 == 1 || n % 12 == 5) {
+		s.prime[n] = !s.prime[n]
+	}
+}
+
+fn (mut s Sieve) step2(x usize, y usize) {
+	n := (3 * x * x) + (y * y)
+	if n <= s.limit && n % 12 == 7 {
+		s.prime[n] = !s.prime[n]
+	}
+}
+
+fn (mut s Sieve) step3(x usize, y usize) {
+	n := (3 * x * x) - (y * y)
+	if x > y && n <= s.limit && n % 12 == 11 {
+		s.prime[n] = !s.prime[n]
+	}
+}
+
+fn (mut s Sieve) loop_y(x usize) {
+	for y := usize(1); y * y < s.limit; y++ {
+		s.step1(x, y)
+		s.step2(x, y)
+		s.step3(x, y)
+	}
+}
+
+fn (mut s Sieve) loop_x() {
+	for x := usize(1); x * x < s.limit; x++ {
+		s.loop_y(x)
+	}
+}
+
+fn (mut s Sieve) calc() {
+	s.loop_x()
+	s.omit_squares()
+}
+
+fn generate_trie(l []usize) Node {
+	mut root := Node{}
+	for el in l {
+		mut head := &root
+		for ch in el.str().runes() {
+			if ch !in head.children {
+				head.children[ch] = &Node{}
+			}
+			head = head.children[ch]
+		}
+		head.terminal = true
+	}
+	return root
+}
+
+struct Pair {
+	Node
+	p_prefix string
+}
+
+fn find(upper_bound usize, prefix int) []int {
+	mut primes := Sieve{upper_bound, []bool{len: int(upper_bound + 1)}}
+	primes.calc()
+	str_prefix := prefix.str()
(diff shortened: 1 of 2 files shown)


primes/primes.v
```diff
@@ -0,0 +1,149 @@
...
+
+struct Sieve {
+mut:
+	limit usize
+	prime []bool
+}
+
+fn (s Sieve) to_list() []usize {
+	mut result := [usize(2), 3]
+	for p := usize(5); p <= s.limit; p++ {
```
dumblob, line 22: Didn't try this, but in V I'd probably write something like `for p in usize(5) .. s.limit +1 {`. But YMMV :wink:.

primes/primes.v
```diff
@@ -0,0 +1,149 @@
...
+		if s.prime[p] {
+			result << p
+		}
+	}
+	return result
+}
+
+fn (mut s Sieve) omit_squares() {
+	for r := 5; r * r < s.limit; r++ {
```
dumblob, line 31: Same here - I'd use the `in` for loop :wink:.

primes/primes.v
```diff
@@ -0,0 +1,149 @@
...
+
+fn (mut s Sieve) step3(x usize, y usize) {
+	n := (3 * x * x) - (y * y)
+	if x > y && n <= s.limit && n % 12 == 11 {
+		s.prime[n] = !s.prime[n]
+	}
+}
+
+fn (mut s Sieve) loop_y(x usize) {
+	for y := usize(1); y * y < s.limit; y++ {
```
dumblob, line 62: Same here...

primes/primes.v
```diff
@@ -0,0 +1,149 @@
...
+		s.step1(x, y)
+		s.step2(x, y)
+		s.step3(x, y)
+	}
+}
+
+fn (mut s Sieve) loop_x() {
+	for x := usize(1); x * x < s.limit; x++ {
```
dumblob, line 70: And here as well.

primes/primes.v
```diff
@@ -0,0 +1,149 @@
...
+	p_prefix string
+}
+
+fn find(upper_bound usize, prefix int) []int {
+	mut primes := Sieve{upper_bound, []bool{len: int(upper_bound + 1)}}
+	primes.calc()
+	str_prefix := prefix.str()
+	mut head := generate_trie(primes.to_list())
+	for ch in str_prefix.bytes() {
+		head = *head.children[ch]
```
dumblob, line 106: The asterisk (`*`) here shouldn't be needed (V does fully automated dereferencing) - actually the compiler should have issued an error about this. Could you please file a bug on this?

primes/primes.v
```diff
@@ -0,0 +1,149 @@
...
+			}
+			head = head.children[ch]
+		}
+		head.terminal = true
+	}
+	return root
+}
+
+struct Pair {
+	Node
```
dumblob, line 96: Did you really intended to embed (i.e. copy) the whole `Node` into Pair? E.g. Zig implementation (the only other implementation I've seen :wink:) doesn't do any copy and instead uses a reference.

primes/primes.v
```diff
@@ -0,0 +1,149 @@
...
+	mut queue := []Pair{}
+	queue.prepend(Pair{head, str_prefix})
+	mut result := []int{}
+	for queue.len > 0 {
+		pair := queue.pop()
+		if pair.terminal {
+			result << pair.p_prefix.int()
+		}
+		for ch, v in pair.children {
+			queue.insert(0, Pair{v, pair.p_prefix + ch.ascii_str()})
```
dumblob, line 117: Here `v` is being dereferenced and then copied I think.

OpenSourceAnarchist:
@dumblob Thank you for all the feedback! As for using C-style loops, I figured using `in` is more idiomatic for V, but it felt weird adding one to `s.limit` and saw it was supported in the documentation. I figured `in` with a range makes more sense when you want a strictly less/greater than scenario rather than less/greater than _or_ equal to, but I see your point.

I didn't know about automatic de-referencing... I'll file a bug report.

As for embedding the struct... I wanted to use a tuple, but V doesn't have them, so I just made a new struct type. I know the long-term goal of V is to have the compiler decide whether to pass a struct by value or by reference, but you're right that it would probably, until that point, to be specific and use a reference. In general, I avoided using references in my code unless absolutely necessary since I figure the point is to let the compiler make the right decision...

Same point for `v` being de-referenced and copied... shouldn't the compiler make that decision rather than the programmer? Or maybe I'm misunderstanding that part of V's goals.

Either way, thank you for all the suggestions! If you think specifying `&Node` and `&v` is important, I'm sure nuald would accept a PR. If `autofree` is enabled, it really shouldn't matter that much, and hopefully the compiler will get smarter with passing by reference or by value.

dumblob:
> @dumblob Thank you for all the feedback! As for using C-style loops, I figured using `in` is more idiomatic for V, but it felt weird adding 1 to `s.limit` and saw it was supported in the documentation. I figured `in` with a range makes more sense when you want a strictly less/greater than scenario rather than less/greater than _or_ equal to, but I see your point.

Sure, it's probably more about the taste. OTOH `in` syntax is more expressive/readable (and a teeny-tiny bit more optimizable) and thus should be more "V-ish" (even at the expense of `+1`) :wink:.

> ~I didn't know about automatic de-referencing... I'll file a bug report.~ I checked on discord, and automatic de-referencing is no longer a planned feature... "too dangerous" :)
>
> As for embedding the struct... I wanted to use a tuple, but V doesn't have them, so I just made a new struct type. I know the long-term goal of V is to have the compiler decide whether to pass a struct by value or by reference, but you're right that it would probably be better, until the compiler's features improve, to be specific and use &Node. In general, I avoided using references in my code unless absolutely necessary since I figure the point is to let the compiler make the right decision...
> 
> Same point for `v` being de-referenced and copied... shouldn't the compiler make that decision rather than the programmer? Or maybe I'm misunderstanding that part of V's goals.
> 
> Either way, thank you for all the suggestions! If you think specifying `&Node` and `&v` is important, I'm sure nuald would accept a PR. If `autofree` is enabled, it really shouldn't matter that much, and hopefully the compiler will get smarter with passing by reference or by value.

Sorry for the misunderstanding here. It's true, that automatic dereferencing of **explicit** references was always a questionable feature of *V* (and I don't follow the super-speedy development of *V* so I'm sorry for suggesting this).

But what's more important is the fact, that automatic dereferencing **is a thing in V** - but for **implicit** references. Which is typically everything `mut` in a struct or a function argument list.

So in other words:
1. `struct S{a Node}` is immutable "embedding" - i.e. always a copy of `Node` (no referencing nor dereferencing here - the compiler is forced by the programmer to **always** copy the data)
2. `struct S{a &Node}` is an **explicit** reference - i.e. guaranteed to **not** copy but instead make a reference (note this gives the compiler **zero** freedom to e.g. copy instead of making a reference if the compiler would think it's faster) - it's discouraged in V to use these references unless absolutely necessary
3. `struct S{mut a Node}` is **implicit** reference which might be implemented however the compiler wants (the programmer has no clue how it's implemented under the hood) - in this case the compiler is totally free to make it reference on one place and an embedding on another - and this is exactly where auto-dereferencing happens

The case (2) is what people on Discord talked about.

So my suggestion is to make the tuple struct `mut` to let the compiler decide what's faster instead of the currently forced copy.

Also `autofree` has nothing to do with this (except maybe the fact that there is this `[heap]` attribute which *V* will demand if its static analysis is unable to determine how to "borrow" the memory reliably - but the compiler will tell you this and won't compile an errorneous code, so don't worry).

Does this make more sense?

dumblob:
>So my suggestion is to make the tuple struct `mut` to let the compiler decide what's faster instead of the currently forced copy.

Or if we want to be "fair" (comparable to Zig), we'd really need to make it an explicit reference. What do you think?

OpenSourceAnarchist:
@dumblob Thanks for the clarifications! I am new to V and find it hard to keep up with all the development discussions, but your explanation about de-referencing makes a lot of sense, and how struct embedding differs.

I tried changing the code from struct embedding of `Node` to using mutable fields in the `Pair` struct, and then tried using standard immutable fields but making it `&Node`, and finally a mutable `&Node` just to see if there was any difference. I ran benchmarks (I cloned this repo and edited the Makefile to just run the V tests) for each variation. From what I can tell, none of these changes made any difference in terms of speed or memory usage. In fact, compiling with `-autofree` made all of them run faster and use _slightly_ (a few Mb) less memory on average, and this required adding `[heap]` above the `Node` struct definition like you said it would (whenever `Pair` had `&Node` as a field).

So, as far as I can see, any of these changes would be cosmetic right now, and (hopefully) potentially useful as V's compiler improves. I have no idea why the memory usage is so large relative to the other compiled languages on the list... feel free to make a PR with your suggested changes (or just reply exactly what changes you'd like) just in case I didn't implement what you wrote above correctly. If we're going to edit the implementation, I would prefer to not use explicit references, just implicit with mutable struct fields.

dumblob:
I actually don't think it'd that necessary to make any changes as of now. I've just commented on things which came to my mind when reading the code and comparing it to Zig :wink:.

Once V will mature (especially when `-autofree` becomes the default), we can revise the current implementation :wink:.

So don't worry and let's leave it as it is.

Btw. we could also say, that because V boasts itself to be easy for newcomers and you (in your own words) were/are a newcomer, then in accordance with the mantra of this benchmark repository, the code you've written is actually "how you'd write it in practice if under time pressure (i.e. without any optimization)".

PR #399, opened 2022-03-13
Optimize the C version of the BF

ricvelozo:
Use setbuf() to use unbuffered output, and memset() to init the tape.

### DIFF
--- a/brainfuck/bf.c
+++ b/brainfuck/bf.c
@@ -116,7 +116,6 @@ void print(struct printer *p, int n) {
     p->sum2 = (p->sum2 + p->sum1) % 255;
   } else {
     putc(n, stdout);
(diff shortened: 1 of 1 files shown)


dumblob:
If this gets merged, then other implementations should also get the `flush()` removed from `print()` methods etc. (e.g. [V](https://github.com/kostya/benchmarks/blob/master/brainfuck/bf.v#L82 )).

ricvelozo:
> If this gets merged, then other implementations should also get the `flush()` removed from `print()` methods etc. (e.g. [V](https://github.com/kostya/benchmarks/blob/master/brainfuck/bf.v#L82 )).

The flush wasn't removed, the code is using `setbuf()`.

PR #402, opened 2022-03-29
Added a Rust benchmark with a crate

betseg:
I copied the Makefile targets from the JSON benchmark into this one.

On my machine the existing program runs around 4.6 seconds, and the one I'm adding runs in around 120 ms.


matmul/rust-ndarray/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.0.1"
+edition = "2021"
+
+[profile.release]
+lto = true
+codegen-units = 1
+panic = "abort"
+
+[dependencies]
+ndarray = "0.15"
```
nuald, line 12: Please use the most specific version as possible, so it's easier to track the changes and versions. Right now, it's "0.15.4".

Another note here is that you use the default feature set that could be not the ideal for performance tests. BLAS integration and other features could be a game changer. I'd leave it to you to experiment here.

Also, please add a newline at the end of the file (`git` usually warns about it too).

matmul/rust-ndarray/src/main.rs
```diff
@@ -0,0 +1,53 @@
...
+    let i = Array::from_shape_fn((n, n), |(i, _)| i as f64);
+    let j = i.t();
+
+    let n = n as f64;
+    (&i - &j) * (&i + &j) * (seed / n / n)
+}
+
+fn mat_mul(a: &Array2<f64>, b: &Array2<f64>) -> Array2<f64> {
+    let mut c = Array2::<f64>::zeros((b.shape()[0], a.shape()[1]));
+    general_mat_mul(1., a, b, 1., &mut c);
```
nuald, line 16: Would it be more idiomatic to use `dot`?

matmul/rust-ndarray/src/main.rs
```diff
@@ -0,0 +1,53 @@
...
+        .unwrap();
+
+    let left = calc(101);
+    let right = -18.67;
+    if (left - right).abs() > 0.1 {
+        eprintln!("{} != {}", left, right);
+        process::exit(-1);
+    }
+
+    notify(&format!("Rust\t{}", process::id()));
```
nuald, line 48: Please specify the library here, I guess it would be "Rust (ndarray)"

betseg: I changed everything and forgot this one.

nuald:
I've added the numbers into README. Also changed code to use multithreaded `matrixmultiply` instead of BLAS - apparently, the latter doesn't use multithreaded mode, therefore the numbers are not so great (twice slower for me). Also BLAS uses twice as much memory too comparing with `matrixmultiply`.

PR #404, opened 2022-03-31
Added Julia implementation for Primes benchmark

lfenzo:
The missing Julia implementation for the "Primes" benchmark has been added.  I wasn't sure how to edit `primes/Makefile` properly so I made no changes to it. If anyone knows how to add this to the current target/run pipeline, please, feel free to make the required changes.


nuald:
Applied all additional changes required for the benchmark, plus added the result numbers to save time later.

lfenzo:
Hi @nuald,

Thanks for adding it to the pipeline! I wasn't sure on adding the results just yet given that it could introduce some testing environment inconsistencies; but in the end of the day these values can be overwritten by any maintenance update so I think that keeping them in the PR is just fine.

I believe this is now ready to merge.

PR #410, opened 2022-05-31
use types properly in Julia implimentation of Primes benchmark

oscardssmith:
I've had to try very hard not to optimize out the stupidness of this benchmark (for example, the `Primes` library in Julia has a fairly simple pure Julia sieve of Eratosthenes  that is 60x faster)
```
using Primes
@time primes(UPPER_BOUND);
  0.008639 seconds (5 allocations: 3.939 MiB)
```
but I believe this is still fair.

### DIFF
--- a/primes/primes.jl
+++ b/primes/primes.jl
@@ -6,7 +6,7 @@ const PREFIX = 32_338
 
 
 mutable struct Node
-    children::Dict
+    children::Dict{Char, Node}
     terminal::Bool
 end
 Node(children::Dict) = Node(children, false)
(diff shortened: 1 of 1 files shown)


nuald:
The benchmark is not about the fastest primes calculation, but about the general performance of working with the various data structures. It's still not fair, but can gives one some insights and it could be a base for the further optimization and tests.

As for Eratosthenes, it's too simple for benchmarking, and in majority of the cases it's about memory access speed. Also, as it's quite straightforward, it could utilize CPU caches and SSE & AVX vectorization, but it's out of scope of this project.

PR #414, opened 2022-06-14
New Mir Ion pasrser implementation 

9il:
Please update the table then. It should be faster then Asdf, I hope so.


json/test_mir_ion_file.d
```diff
@@ -1,18 +0,0 @@
...
-dflags "-mcpu=native" "-linkonce-templates" "-enable-cross-module-inlining" platform="ldc"
-+/
-import mir_common_json;
-import mir.stdio;
-import mir.deser.json : deserializeJsonFile;
-
-void main() @safe @nogc
-{
-    "D/ldc2 (Mir Amazon's Ion DOM, file input)".notifyStart;
-    auto coordinate = "/tmp/1.json".deserializeJsonFile!Avg.coordinates.avg;
```
9il, line 15: The function is no longer supported.

json/mir-common-json/source/mir_common_json.d
```diff
@@ -49,6 +49,7 @@ auto notifyStop(scope const char[] msg = "stop")
     notify(msg);
 }
 
+@serdeIgnoreUnexpectedKeys
```
nuald, line 52: Could you please clarify why did you add this line (and the same below)?

9il: Unlike the old one, the new implementation will throw exceptions in case of keys that hasn't been handled by the structure. The UDAs are used to preserve the old behaviour and do not throw.

9il: For example, do not throw exception, that this input isn't used:
```
      "name": "wxetbf 1966",
      "opts": {
        "1": [
          1,
          true
        ]
      }
```

nuald:
Unfortunately, the newer version is even slower. Please ensure that you use the same LDC version (1.29.0) - I've already seen the cases when the compiler upgrade regressed the tests performance.

9il:
Haha, how bad it is?

nuald:
Please see #415 (new 231ms vs old 192ms).

PR #416, opened 2022-06-15
Simplified multiplication of the matrices

ColourGrey:
The version that I am modifying created a transpose `c` of matrix `b`, which is completely unnecessary. This meant wasted memory and wasted CPU cycles, amounting to a program slower than it should have been for such an elementary task (in particular, it was slower than almost half of the other benchmarked programs written in the other languages). The version that I am proposing is functionally identical, but uses less memory and is faster by dropping the needless intermediate allocations for the transpose.

### DIFF
--- a/matmul/matmul.go
+++ b/matmul/matmul.go
@@ -29,25 +29,16 @@ func matmul(a [][]float64, b [][]float64) [][]float64 {
 	m := len(a)
 	n := len(a[0])
 	p := len(b[0])
-	x := make([][]float64, m)
-	c := make([][]float64, p)
-	for i := 0; i < p; i++ {
-		c[i] = make([]float64, n)
-		for j := 0; j < n; j++ {
-			c[i][j] = b[j][i]
-		}
-	}
-	for i, am := range a {
-		x[i] = make([]float64, p)
-		for j, cm := range c {
-			s := float64(0)
-			for k, m := range am {
-				s += m * cm[k]
+	c := make([][]float64, m)
+	for i := 0; i < m; i++ {
+		c[i] = make([]float64, p)
+		for j := 0; j < p; j++ {
+			for k := 0; k < n; k++ {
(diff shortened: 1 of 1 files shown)


nuald:
Transpose solution is more CPU cache friendly. I've tested your change and applied the similar change to C test (to avoid any GC issues), and the "cleaner" solution doesn't perform better (but surely consume less memory):

- Go: with transpose 3.149 seconds and 76 Mb, without transpose 16.112 seconds and 58 Mb
- C/gcc: with transpose 3.028 seconds and 70 Mb, without transpose 4.9 seconds and 53 Mb

I guess you may get different results on desktop CPU which usually have smaller CPU caches comparing with the server ones. Standard implementations (like OpenBLAS) are also heavily optimized to avoid any CPU cache misses. I've stumbled upon this article: https://attractivechaos.wordpress.com/2016/08/28/optimizing-matrix-multiplication/ You may find it interesting as it shows the different techniques and refers to the source code you may play with if you want.

nuald:
You're absolutely right, however let me please notice that using libraries/variances is fine if the reference implementation is presented. Essentially, there are 3 areas of improvement here, and it would be nice if you open the tickets to track it (I can't guarantee that I'll have time in near future to address it though, but all contributions are welcomed):

1. All reference implementations should use the same algorithms (i.e. multiplication with transposing). As you've noted, matmul.jl is not correct. matmul-native.jl could be considered as a variance and allowed to be used.
2. All variances or library-based tests should have the reference implementation. java-nd4j is fine as we have matmul.java, rust-ndarray is fine as we have matmul.rs, matmul-numpy is fine as we have matmul.py. C++ and Tcl are violations as we don't have the reference implementations.
3. The library-based tests and variances should be in other tables (it's not about matmul only, but other tests too). It will help to avoid confusion during results comparison.

PR #420, opened 2022-06-21
brainfuck: minor improvements to V implementation

hungrybluedev:
## Summary

- The StringIterator was almost perfect. Changed it slightly to make the code more idiomatic.
- Use enums instead of consts.
- `byte` to `u8` conversion with `v fmt`

## Goals

This implementation is slightly faster than the older one. At least on my machine (Ryzen 5 4500U). It should bring the V/gcc and V/clang times closer to the C versions for both `bench.b` and `mandel.b`.

### DIFF
--- a/brainfuck/bf.v
+++ b/brainfuck/bf.v
@@ -1,27 +1,27 @@
 import os
 import net
 
-const (
-	inc   = 0
-	move  = 1
-	print = 2
-	loop  = 3
-)
+enum OpType {
+	inc
+	move
+	print
+	loop
+}
 
 struct Op {
-	o    int
+	o    OpType
 	v    int
 	loop []Op
 }
 
-fn new_op(op int, v int) Op {
+fn new_op(op OpType, v int) Op {
 	return Op{
 		o: op
 		v: v
 	}
 }
 
-fn new_op_loop(op int, l []Op) Op {
+fn new_op_loop(op OpType, l []Op) Op {
 	return Op{
 		o: op
 		loop: l
@@ -52,16 +52,16 @@ fn (mut t Tape) inc(x int) {
 
 fn (mut t Tape) move(x int) {
 	t.pos += x
-	for t.pos >= t.tape.len {
-		t.tape << 0
+	if t.pos >= t.tape.len {
+		t.tape << [0].repeat(t.pos - t.tape.len + 1)
 	}
 }
 
 struct Printer {
 	quiet bool
 mut:
-	sum1 u32
-	sum2 u32
+	sum1 u8
+	sum2 u8
 }
 
 fn new_printer(quiet bool) Printer {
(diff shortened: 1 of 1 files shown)


brainfuck/bf.v
```diff
@@ -52,16 +52,16 @@ fn (mut t Tape) inc(x int) {
...
 	}
 }
 
 struct Printer {
 	quiet bool
 mut:
-	sum1 u32
-	sum2 u32
+	sum1 u8
+	sum2 u8
```
dumblob, line 64: I did not examine other languages than Go and C but at least with these two this is **very** unfair.

brainfuck/bf.v
```diff
@@ -175,13 +172,13 @@ fn new_si(s string) StringIterator {
 	}
 }
 
-fn (mut si StringIterator) next() byte {
+fn (mut si StringIterator) next() ?u8 {
```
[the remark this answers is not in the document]
hungrybluedev: The documentation on iterators still show using an optional with an error. Maybe that needs an update. Will test suggestion and report back.

brainfuck/bf.v
```diff
@@ -52,16 +52,16 @@ fn (mut t Tape) inc(x int) {
 
 fn (mut t Tape) move(x int) {
 	t.pos += x
-	for t.pos >= t.tape.len {
-		t.tape << 0
+	if t.pos >= t.tape.len {
+		t.tape << [0].repeat(t.pos - t.tape.len + 1)
```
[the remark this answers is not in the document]
hungrybluedev: There is a grow to length function but it's unsafe. Also the C and Java versions seem to be just doubling capacity when position exceeds tape length. This behaviour, at least, is not standardised.

I'll see if there's a more elegant solution.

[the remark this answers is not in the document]
dumblob: Do you mean grow to length in the `repeat()` method? I actually meant something as simple as instead of the current
```v
if t.pos >= t.tape.len {
		t.tape << [0].repeat(t.pos - t.tape.len + 1)
}
```
do just something like
```v
t.tape << [0].repeat(t.pos - t.tape.len + 1)
```
(`[0].repeat(0)` does return an empty array)

spytheman, line 56: What is weird, is that the grow is always 1, for all brainfuck programs that I tried, i.e. `t.tape << 0` should have been more efficient compared to creating a new array with repeat, and pushing it 🤔 , but it is not.

Update: 
just `if t.pos >= t.tape.len { t.tape << [0] }` is more efficient 
than `if t.pos >= t.tape.len { t.tape << 0 }` too, which is also weird.

spytheman: I think, that `for t.pos >= t.tape.len { t.tape << [0] }` is the most idiomatic, while still faster than the current one, with the current implementation of V.

spytheman: It turns out, that `t.tape << [0]` being faster than `t.tape << 0`, is a side effect of the garbage collector, that is now on by default with V 0.3; with `-gc none`, it is the reverse.

Issue #312, opened 2021-03-15
Matmul tests don't seem to be representative (Julia vs Python experiments)


no-trick-pony:
The following tests were performed with:

Python 3.8.2 + numpy 1.20.1
Julia 1.5.3
Windows 10
Ryzen 7 1700

I am currently trying to figure out which language to use for my project which will include a lot of smaller matrix multiplications. I looked at Rust, Julia as well as Python. According to the Matmul benchmark here, Julia-native is twice as slow as Python-numpy with very little variance. That seemed kinda odd, since Julia was designed to be used in the scientific realm/Computer Science, where one would assume that linear algebra is a foundation for many things - if they found Julia to be that much slower, they surely would have opted to have a look at numpy and improve their own implementation. It's probably the one thing Julia should be somewhat good at. ^^ I tested the two scripts (which I modified to output on stdout) but had about the same results. But only for a very small matrix sizes and mismatched thread count.

###
Experiments:

So I tried it with 10000 as a parameter for the matrix sizing instead:
- Julia used 50% CPU, 8 threads and took 17.5 seconds on average
- Numpy used 100% CPU, (I assume) 16 threads and took 15.5 seconds on average

Hm. Why Julia is using 8 threads on my 16 thread CPU.. I have no idea and it should maybe be changed by the Julia team as well as the fact that BLAS is ignoring the `--thread` option and the thread environment variable but instead has to be manually set in code or the specific BLAS environment variable.

I changed the code to allow to specify the number of threads and tested again, for 4, 8 and 16 threads each (10000 again as a param). Note, that I ran every test only once. After the change to `calc(n + (i % 2))` (see below) I reran the tests but the outcome was roughly the same in terms of relative time difference. The code in the brackets is how you would call the final code down below - parameters are `matrix sizing`, `number of iterations` and `number of threads to use`:
- Julia, 4 threads: 30.9 s (`julia matmul-native.jl 10000 1 4`)
- Numpy , 4 threads: 32.67s (`python matmul-numpy.py 10000 1 4`)
- Julia, 8 threads: 17.7s (`julia matmul-native.jl 10000 1 8`)
- Numpy , 8 threads: 18.6s (`python matmul-numpy.py 10000 1 8`)
- Julia, 16 threads: 13.8s (`julia matmul-native.jl 10000 1 16`)
- Python, 16 threads: 15.7s (`python matmul-numpy.py 10000 1 16`)

Then I looked into the Matmul Makefile and found out that 1500 was used there:
[...]

PR #422, opened 2022-08-15
Optimize the V version of the BF interpreter.

spytheman:
V does bounds checking by default on each array access. 

Using `[direct_array_access]` over a function/method, disables the checks inside its body.

The brainfuck implementation does call very frequently the methods .get() and .inc() , which boil down to array access operations.

With this change, the V implementation is within 5% of the C version, when it is compiled with `-prod` (and the C version is compiled with -O3) on my machine.

### DIFF
--- a/brainfuck/bf.v
+++ b/brainfuck/bf.v
@@ -42,10 +42,12 @@ fn new_tape() Tape {
 	return t
 }
 
+[direct_array_access]
 fn (t Tape) get() int {
 	return t.tape[t.pos]
 }
 
+[direct_array_access]
 fn (mut t Tape) inc(x int) {
 	t.tape[t.pos] += x
 }
--- a/brainfuck/bf.v
+++ b/brainfuck/bf.v
@@ -55,7 +55,7 @@ fn (mut t Tape) inc(x int) {
 fn (mut t Tape) move(x int) {
 	t.pos += x
 	for t.pos >= t.tape.len {
-		t.tape << 0
+		t.tape << [0]
 	}
 }
 


dumblob:
Related: https://github.com/kostya/benchmarks/issues/312#issuecomment-918371330

brainfuck/bf.v
```diff
@@ -42,18 +42,20 @@ fn new_tape() Tape {
...
+[direct_array_access]
 fn (mut t Tape) inc(x int) {
 	t.tape[t.pos] += x
 }
 
 fn (mut t Tape) move(x int) {
 	t.pos += x
 	for t.pos >= t.tape.len {
-		t.tape << 0
+		t.tape << [0]
```
spytheman, line 58: After V 0.3, the garbage collector is on by default, and pushing [0] is faster than just 0.

nuald:
Sorry, those changes are micro-optimizations, and ideally could be provided by the compiler itself:

- `[direct_array_access]` could be enabled for all array operations if V lang would have the special compiler flag for that (like `danger` in Nim: https://nim-lang.org/docs/nimc.html#additional-compilation-switches)
- appending an array vs appending an element performance seems like a compiler issue too as it's not an expected behaviour.

Please note that we don't compare top performance here, but rather the performance of the code written by the average software developer, and it's unlikely that they use constructions like `[direct_array_access]` in the usual code.

ylluminate:
These constructs are actually encouraged and shared to regular developers @nuald. Please feel free to pop on into Discord for further discussion: https://discord.gg/vlang

spytheman:
bump

spytheman:
May be obsoleted by <https://github.com/kostya/benchmarks/pull/431> , which will also benefit other V programs in this benchmark suite.

PR #423, opened 2022-08-18
Added a useful link

Vaishnaviamirayada:
Have added a useful link about the parsing JSON file from the scaler topics.


kostya:
spam

PR #427, opened 2022-10-16
Optimize the C++ version of Primes

jonathanpoelen:
First commit:
- Replace `shared_ptr` with `unique_ptr` which makes more sense for pointer lifetime.
- `Node`s are allocaed only if the key is missing from `std::unordered_map`.

Second commit:
- Use an arena allocator as Zig


nuald:
I doubt that using arena allocator is a typical use-case for C++ (at the end, the new/delete behaviour could be changed even in runtime, not for Zig though unless it uses the appropriate allocator). However, it seems fair enough as it doesn't look like micro-optimization, but rather proper code for constrained environments. My only concern is about `emplace` as using `find` before it seems redundant. You may look into implementation in LLVM if you don't trust documentation: https://github.com/llvm/llvm-project/blob/d7d586e5a7d71082200231a7cd407cae50924f9c/libcxx/include/__hash_table#L1994

jonathanpoelen:
> My only concern is about emplace as using find before it seems redundant.

Yes... But what is expensive is the dynamic allocation:

```cpp
auto it = head->children.find(ch);
if (it == head->children.end()) {
    it = head->children.emplace(ch, make_node(alloc)).first;
}
```
vs
```cpp
auto it = head->children.emplace(ch, make_node(alloc)).first;
```

algo | time | memory
:---: | :---: | :---:
no find | 0.17s | 257264K
with find | 0.10s | 94356K

(This was done in java and zig, so i tried that too)

nuald:
Right, it's not a lazy initialization, I've missed it. And you're right that C++ implementation is different from the others here.

I have to note though that it's still not efficient enough as it does the double lookup anyway (`find` and then `emplace` even if we know that it's not going to find anything). It's discussed and the solution proposed in https://jguegant.github.io/blogs/tech/performing-try-emplace.html I don't think it would be the micro-optimization as other implementation have it anyway (like Rust with [or_insert_with](https://doc.rust-lang.org/stable/std/collections/hash_map/enum.Entry.html#method.or_insert_with) ) so it could be okay to introduce a little improvement here if you want.

jonathanpoelen:
I haven't noticed any difference in speed with my versions of gcc and clang.

However, I realize that `<memory_resources>` and the whole `std::pmr` namespace is only available since clang-16 for libc++... The readme mentions clang-14.

Doing the equivalent of a minimal `std::pmr::monotonic_buffer_resource` should only take about 30 lines, but wouldn't that be a bit too specific?

nuald:
Debian usually have pretty much old packages even in testing. I'll try to change the image to use newer LLVM (https://apt.llvm.org), no need to use workarounds.

nuald:
I haven't realized that clang-16 is in the pre-release stage. The installation of development version of LLVM would be quite cumbersome, could you please add those 30 lines with the `TODO` note to remove it when clang-16 is released? Thanks in advance!

jonathanpoelen:
In the first commit I added the classes that replace `std::pmr::monotonic_buffer_resource` and the allocator used by `std::pmr::unordered_map`.

In the second one I replaced `std::unordered_map` by an array of 10 elements (as the keys are digits...) It's 2 times faster, but very specialized. Even if there are actually very few lines modified, to do it well, it should be done with all languages where it is faster.

nuald:
Please rebase the changes first as the code has conflicts which can't be resolved automatically. Second, I have few comments I'd like to be addressed:

- using `std::array` is a micro-optimization (it's totally correct, but the goal is comparing performance of standard containers in various programming languages). It's fine to have the code as the second, optimized, version, but the reference implementation should use the hashmap. I guess the proper way would be modifying the tests to make it impossible to make that optimization, but it's out of scope of this change;
- please put the proper guards to use an available `pmr`, as now if any C++ implementation would start providing it, we won't notice it as the newly introduced allocator is always used;
- there are some magic numbers in the custom allocator. I guess that those numbers are always valid in x86_64 Linux, but I'd prefer to have the proper constants in place as the code is copy-pasted to different platforms too.

jonathanpoelen:
I added a condition with `#if __has_include(<memory_resource>)` to use `<memory_resource>` if available, made `monotonic_buffer_resource` less dependent on the requested size (it didn't work if the requested size was bigger than the minimum buffer size) and replaced the magic value with a constant.

PR #428, opened 2022-10-20
bf.swift: use unsafe buffer pointer to the tape

ricvelozo:
Small optimization similar to the Rust version (7e2fc03).

### DIFF
--- a/brainfuck/bf.swift
+++ b/brainfuck/bf.swift
@@ -16,8 +16,8 @@ struct Tape {
     var pos = 0
     var tape: ContiguousArray<Int32> = [0]
     var currentCell: Int32 {
-        get { tape[pos] }
-        set { tape[pos] = newValue }
+        get { tape.withUnsafeBufferPointer { $0[pos] } }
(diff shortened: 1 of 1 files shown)


nuald:
The numbers are up-to-dated. Please note that the changes are negligible because we use `-Ounchecked` compiler option for Swift.

kostya:
is not this is the same microoptimization like this: https://github.com/kostya/benchmarks/pull/422/files#diff-da22c1257fb473df31008d86b02a973793bdeefe14abe5e9de076a782475868dR45

nuald:
Almost, and if the aforementioned change would only have `[direct_array_access]` added, I think we could merge it (as this approach could be definitely seen in average code). However, another change (`t.tape << [0]`) is definitely a compiler issue as it's not expected that pushing an array is faster than pushing one element.

ricvelozo:
> The numbers are up-to-dated. Please note that the changes are negligible because we use `-Ounchecked` compiler option for Swift.

I think `-Ounchecked` just disables integer overflow checks.

PR #431, opened 2022-10-30
Update V_FLAGS to use the COMMON_FLAGS too + the new -no-bounds-checking option

spytheman:
Since today (from commit 9edb485), the main V compiler implementation, supports
a new option `-no-bounds-checking` . This PR updates V_FLAGS to use
it, and also to pass the COMMON_FLAGS to the C backend, just like
compilers for other languages do.


spytheman:
@nuald, @kostya Could you also please do a followup on this PR, by updating the V results in the README.md, just like you did for #438 ?

Many thanks in advance.

spytheman:
Thank you!

PR #432, opened 2022-11-01
Add naive pure functional Idris variant

philnguyen:
This naive functional implementation (using linked lists) in a [research-quality language](https://www.idris-lang.org) actually runs pretty competitively with the production-quality ones. Let there be the first benchmark submission in a dependently-typed language ^_^ (although the implementation doesn't use any. It abstracts over the execution context `Loud` or `Quiet`, but technically the Haskell implementation could have done that too).

I don't know all the steps to add a new language to the benchmark suite, but I can list the commands to build and run below, and hopefully somebody is interested in helping me?

To install Idris2:
* Download from [https://www.idris-lang.org/pages/download.html](https://www.idris-lang.org/pages/download.html)
* Assume Chez Scheme installed
* Edit `PREFIX` in `config.mk` or leave it default `$HOME/.idris2`. Make sure `$PREFIX/bin` is in `$PATH` and `$PREFIX/lib`  is in `$LD_LIBRARY_PATH` (or `$DYLD_LIBRARY_PATH` on mac)
* `make bootstrap SCHEME=chez && make install && make all && make install`

To build `bf.idr`:
* `idris2 -p network bf.idr -o bf_idris`

To run the built executable on `bench.b`:
* `./build/exec/bf_idris bench.b`

### DIFF
--- a/brainfuck/bf.idr
+++ b/brainfuck/bf.idr
@@ -1,0 +1,118 @@
+module Main
+
+import Control.Monad.State as St
+import Data.String as Str
+import Data.Bits as B
+import Network.Socket as Skt
+import System as Sys
+import System.File as F
+
+data Op = Inc Int | Left | Right | Print | Loop (List Op)
+data Tape = MkTape (List Int) Int (List Int)
+
+current : Tape -> Int
+current (MkTape _ n _) = n
+
+inc : Int -> Tape -> Tape
+inc d (MkTape l n r) = MkTape l (n + d) r
+
+decomp : List Int -> (List Int, Int)
+decomp (x :: xs) = (xs, x)
+decomp [] = ([], 0)
+
+left, right: Tape -> Tape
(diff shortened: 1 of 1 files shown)


nuald:
I'll add the necessary components to make it work.

nuald:
Added the toolchain and test in #437 The local run shows that its performance is on par with Haskell. Also I've fixed few IO in the code to be properly execute in our environment. The actual numbers will be posted later within the maintenance update soon.

nuald:
Could you please fix the quiet mode? In #441 I've disabled it as the operation sequence wasn't correct - it should:

1. Notify the server with the test name
2. Run calculation
3. Notify the server to stop

If you have Ruby installed, you could verify it by running:

```sh
idris2 -p network bf.idr --output-dir target -o bin_idris
QUIET=1 ../xtime.rb target/bin_idris bench.b
```
It should properly report the number of seconds the test takes.

philnguyen:
I followed up in #442. Thanks!!

PR #435, opened 2022-11-02
Modernize Rust code

ricvelozo:
* Use captured identifiers in format strings
* Use raw strings for JSON
* Break big string in lines with `concat!()`
* Use a tuple to represent a socket address
* Don't parse default value of env arg

### DIFF
--- a/base64/base64.rs/src/bin/base64.rs
+++ b/base64/base64.rs/src/bin/base64.rs
@@ -7,7 +7,7 @@ const STR_SIZE: usize = 131_072;
 const TRIES: usize = 8192;
 
 fn notify(msg: &str) {
-    if let Ok(mut stream) = TcpStream::connect("localhost:9001") {
+    if let Ok(mut stream) = TcpStream::connect(("127.0.0.1", 9001)) {
         stream.write_all(msg.as_bytes()).unwrap();
     }
 }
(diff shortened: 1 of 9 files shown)


nuald:
Thanks for the contribution! There is another issue that bother me with the Rust code though - abusing `unwrap()`. We should have the proper error propagation with the exit at `main` function (https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html), e.g.

```rs
fn main() -> Result<(), Box<dyn Error>> {
    run_test_code()?;

    Ok(())
}
```

As the code in this repository is used as a sample for other projects, it would be nice if the code is written with the proper error handling. However, it's a little bit time-consuming task, so please let me know if you don't want to take on it (or address later outside of the PR).

ricvelozo:
It's better in another PR.

But, the codes needs to be equivalent, and the C++ version doesn't do checks in many places. Also, `.expect()`/`.unwrap()` is better if you want to check for errors, instead of just return with `?` (the same as `.unwrap()`).

Summary:

- `.unwrap()` - panics with default excuse
- `.expect()` - panics with custom excuse
- `?` - returns and groups errors to handle it in the call site; if not handled, panics in the end

nuald:
C++ error handling could be improved too, but I wouldn't say it doesn't do checks in many places (plus they have exceptions which propagated automatically). For example, `read_file` in C++ checks for file stream, and if it's in the failed state, it returns empty string (I'm not saying it's a correct behaviour, that's just what we have now). Rust version just panic.

The reason I'd prefer to have it in `main` is because the code is copy-pasted, and all those `unwrap()` should be replaced with the proper error propagating sometimes. However, I've grepped the code right now, it's not that many places, so I think it's fine.

PR #438, opened 2022-11-25
Use a more functional implementation for Haskell's tape

philnguyen:
The existing Haskell implementation is trying to emulate the imperative solution. Given that we already have an imperative variant using `MArray`, I propose we have the default implementation use a more FP-friendly data structure based on linked lists (similar to the Idris variant).

On my computer, this implementation speeds up the benchmarks by 2x. This also agrees with the common experience among Haskell people: when a Haskell program is slow, it's usually counter-productive to rewrite it to be more like C.

But I don't have a strong opinion on the naming, and if it's weird to suddenly change the implementation of an established variant, I'm ok with making this a new variant `Haskell (FP)` or something.

### DIFF
--- a/brainfuck/bf.hs
+++ b/brainfuck/bf.hs
@@ -3,8 +3,6 @@
 
 module Main where
 
-import qualified Data.Array.Base as ArrayBase
-import qualified Data.Array.Unboxed as UArray
 import qualified Data.ByteString.Char8 as C
 import Control.Exception
 import Control.Monad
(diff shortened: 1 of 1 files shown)


nuald:
I agree that the current implementation doesn't seem like the usual Haskell code. I doubt that we need the new change as a new variant though. We may use the similar approach as with Tcl (no default test, but FP and OOP variants only). In that case, the only addition would be renaming the test:

```hs
notify $ "Haskell (FP)\t" ++ show pid
```

philnguyen:
I update the `notify` line to `"Haskell (FP)\t"` as suggested. Many thanks!!

nuald:
Updated in #441

PR #441, opened 2022-12-10
Maintenance update 10-Dec-2022

nuald:


README.md
```diff
@@ -169,45 +172,45 @@ Testing base64 encoding/decoding of the large blob into the newly allocated buff
...
-|                     Julia |  5.023<sub>±0.004</sub> |  244.56<sub>±00.09</sub> + 42.28<sub>±00.09</sub> | 188.22<sub>±02.10</sub> |
-|  Ruby/truffleruby (--jvm) |  5.072<sub>±0.012</sub> | 347.27<sub>±07.42</sub> + 168.12<sub>±27.78</sub> | 234.49<sub>±01.53</sub> |
-|                Ruby/jruby | 10.519<sub>±0.026</sub> |  182.81<sub>±02.29</sub> + 80.63<sub>±02.65</sub> | 420.63<sub>±01.57</sub> |
-| Perl (MIME::Base64::Perl) | 13.790<sub>±0.053</sub> |    16.12<sub>±00.07</sub> + 0.31<sub>±00.04</sub> | 573.08<sub>±04.60</sub> |
-|          Ruby/truffleruby | 16.232<sub>±0.006</sub> | 295.77<sub>±00.95</sub> + 232.04<sub>±04.13</sub> | 647.25<sub>±03.84</sub> |
+|            C/gcc (aklomp) |  0.099<sub>±0.000</sub> |     2.07<sub>±00.02</sub> + 0.00<sub>±00.00</sub> |   4.64<sub>±00.06</sub> |
+|          C/clang (aklomp) |  0.099<sub>±0.001</sub> |     2.07<sub>±00.04</sub> + 0.00<sub>±00.00</sub> |   4.59<sub>±00.07</sub> |
+|                       PHP |  0.106<sub>±0.000</sub> |    17.31<sub>±00.09</sub> + 0.00<sub>±00.00</sub> |   4.78<sub>±00.07</sub> |
```
kostya, line 177: php what? they implement aklomp algorithm?

README.md
```diff
@@ -372,28 +375,28 @@ Notes:
...
-|              Python/pypy | 0.880<sub>±0.002</sub> |  63.32<sub>±00.09</sub> + 250.27<sub>±00.06</sub> |  34.53<sub>±00.42</sub> |
-| Ruby/truffleruby (--jvm) | 1.381<sub>±0.026</sub> | 354.60<sub>±10.97</sub> + 552.99<sub>±34.23</sub> |  90.57<sub>±01.36</sub> |
-|             Ruby (--jit) | 1.442<sub>±0.004</sub> | 270.43<sub>±00.06</sub> + 147.02<sub>±00.05</sub> |  59.14<sub>±00.99</sub> |
-|                      Lua | 1.478<sub>±0.004</sub> |   2.28<sub>±00.01</sub> + 283.04<sub>±00.48</sub> |  57.46<sub>±00.33</sub> |
-|         Ruby/truffleruby | 1.506<sub>±0.007</sub> | 294.80<sub>±00.82</sub> + 410.90<sub>±26.05</sub> |  84.20<sub>±00.88</sub> |
-|               Ruby/jruby | 1.990<sub>±0.057</sub> | 182.28<sub>±01.67</sub> + 538.37<sub>±28.70</sub> | 102.14<sub>±03.48</sub> |
-|                     Ruby | 2.057<sub>±0.005</sub> |  13.87<sub>±00.07</sub> + 147.07<sub>±00.04</sub> |  82.80<sub>±00.47</sub> |
-|                   Python | 5.022<sub>±0.024</sub> |  10.49<sub>±00.04</sub> + 234.72<sub>±00.26</sub> | 196.62<sub>±02.52</sub> |
+|                  C++/g++ | 0.071<sub>±0.000</sub> |    3.68<sub>±00.04</sub> + 77.91<sub>±00.26</sub> |   2.71<sub>±00.02</sub> |
```
kostya, line 378: wonder why c++ became much faster

nuald: C++ implementation has changed to use the arena allocator (#427). I don't consider it as a micro-optimization as Zig uses the similar allocator too, and it's a totally valid and proper use-case for the embedded systems (where C++ is used a lot).

It's definitely controversial, but we can't have the perfect similarity anyway, and I don't oppose to use the strong sides of the programming languages.

README.md
dumblob: Hm, `V` has `-prealloc` argument which turns on arena allocation. Other languages have such flags perhaps too. Shouldn't that be considered here if C++ and Zig have already gotten this "privilege"?

Ping @spytheman .

README.md
dumblob: Hm, `V` has `-prealloc` argument which turns on arena allocation. Other languages have such flags perhaps too. Shouldn't that be considered here if C++ and Zig have already gotten this "privilege"?

Ping @spytheman .

nuald: Thanks for the suggestion, it's definitely a useful feature. And while other languages may have it too, I'm not aware of it (including V until you mentioned it, as `v help build` doesn't describe this option). I've taken a look into V source code, apparently it preallocates in 16 Mb chunks that is not that suitable for the embedded systems, however in the future it could be tuned I guess. Nonetheless, I'm fine with using this option, and will provide up-to-dated results soon.

nuald: Addressed in #444.

dumblob: Hm, I meant only in this benchmark. Or do Zig and C++ use arena allocator for everything?

nuald: It depends, there are some inconsistencies, and I don't have much time to address everything. But in reality it doesn't matter much - as you could see in the aforementioned PR, only Primes test is really affected.

PR #442, opened 2022-12-11
Temporary fix for delayed server notification on quiet mode

philnguyen:
For some reason, in quiet mode, the `notify $ "Idris"` line is delayed till the end, after the execution is complete, causing a reported 0s time. Adding `pure ()` at the end mysteriously forces it to be run at the expected time. This is possibly a compiler bug to be fixed later.

Tagging @edwinb in case he has insight into what's going on and if it's a bug that needs fixing.

### DIFF
--- a/brainfuck/bf.idr
+++ b/brainfuck/bf.idr
@@ -97,7 +97,7 @@ notify : String -> IO ()
 notify msg = do
   Right skt <- socket AF_INET Stream 0
   _ <- connect skt (Hostname "localhost") 9001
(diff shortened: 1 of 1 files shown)


nuald:
Updated numbers in #443. Please note that the quiet mode operates slower than the usual mode (I guess it's due to the numbers are passed around, not contained in the Printer object like in Haskell.

philnguyen:
Thanks! That's good for monitoring future improvements to the compiler anyways!

PR #444, opened 2022-12-14
Added -prealloc to V_FLAGS.

nuald:


spytheman:
`-prealloc` is an option, that would not be widely used by most people. It works by using a bump memory allocator instead of the malloc/free provided by the OS or the GC. In effect, it turns off the garbage collector, since from its point of view, everything is allocated in a few large blocks of memory, with many pointers inside them. 

Imho `-prealloc` is only useful for speeding up batch programs, that do many small and frequent allocations, then free everything at the end, like the V compiler itself.

PR #446, opened 2022-12-21
Add matmul benchmarks for V with VSL

ulises-jeremias:
- Added matmul using VSL with:
  - with GCC and Clang
  - with pure V backend and CBLAS


matmul/matmul-vsl.v
```diff
@@ -0,0 +1,59 @@
+import os
+import math
+import net
+import vsl.la
+
+fn matgen(n int, seed f64) &la.Matrix[f64] {
```
penguindark, line 6: we can try with this code if the function is used often:
```v
fn matgen(n int, seed f64)  &la.Matrix[f64] {
    n2 := n * n
    mut a := []f64{len: n2}
    tmp := seed / n2
    mut c := 0
    for i in 0 .. n {
        for j in 0 .. n {
	    v := tmp * f64(i - j) * f64(i + j)
            a[c] = v
            c++
        }
    }
    return la.matrix_raw(n, n, a)
}
```

ulises-jeremias: Applied your suggestion to the code!

PR #448, opened 2023-01-06
Introduce JSON yajl in C test

matwey:


json/test_yajl.c
```diff
@@ -0,0 +1,184 @@
...
+	const char test_buffer_1[] = "{\"coordinates\":[{\"x\":2.0,\"y\":0.5,\"z\":0.25}]}";
+	const char test_buffer_2[] = "{\"coordinates\":[{\"y\":0.5,\"x\":2.0,\"z\":0.25}]}";
+
+	context_t ctx = EMPTY_CTX;
+	yajl_handle yajl = yajl_alloc(&ycallbacks, NULL, &ctx);
+
+	yajl_parse(yajl, (const unsigned char*)test_buffer_1, sizeof(test_buffer_1) / sizeof(test_buffer_1[0]));
+	yajl_complete_parse(yajl);
+
+	if (!(ctx.c.x == 2.0 && ctx.c.y == 0.5 && ctx.c.z == 0.25)) {
```
nuald, line 140: Please declare another const context to compare against, and use its values, e.g.

```c
const context_t expected = {
  .c = {
    .x = 2.0,
   ...
  }
};
...
if (ctx.c.c != expected.c.y || ...)
```

json/test_yajl.c
```diff
@@ -0,0 +1,184 @@
...
+
+typedef struct coordinate_s coordinate_t;
+
+struct context_s {
+	size_t depth;
+	size_t len;
+	char item;
+	char in_coord : 1;
+
+	coordinate_t c;
```
nuald, line 40: Please use more readable name.

json/test_yajl.c
```diff
@@ -0,0 +1,184 @@
...
+		fprintf(stderr, "Cannot open /tmp/1.json: %s\n", strerror(errno));
+		exit(EXIT_FAILURE);
+	}
+
+	ctx = EMPTY_CTX;
+	yajl = yajl_alloc(&ycallbacks, NULL, &ctx);
+
+	notify_with_pid("C/" COMPILER " (yajl)");
+
+	while ((ret = read(fd, buf, sizeof(buf)/sizeof(buf[0])))) {
```
nuald, line 173: All tests parse already loaded buffer in the memory. Please change the logic to load the content first, and then parse it.

json/test_yajl.c
```diff
@@ -0,0 +1,184 @@
...
+	&cb_map_key,
+	&cb_end_map,
+	NULL, /* start_array */
+	&cb_end_array
+};
+
+#define EMPTY_CTX ((context_t){0, 0, 0, 0, {0.0, 0.0, 0.0}})
+
+int main() {
+	const char test_buffer_1[] = "{\"coordinates\":[{\"x\":2.0,\"y\":0.5,\"z\":0.25}]}";
```
nuald, line 131: Please extract the verification functionality into another function.

nuald (changes requested):
On the global note: https://github.com/lloyd/yajl is last updated 7 years ago (however, it's still included to Debian). Do you think it is still worth to use it nowadays?

json/test_yajl.c
matwey: Hello.

While I agree with other comments, I have a question here. `yajl` is streaming parsing library, so if we load all the data into the memory we (unnecessarily) increase the test memory consumption. Should I make two flavors of this test? With the loaded buffer parsing and with streaming reading.

matwey:
> On the global note: https://github.com/lloyd/yajl is last updated 7 years ago (however, it's still included to Debian). Do you think it is still worth to use it nowadays?

Yes, it is included not only in Debian, but in all major Linux distros. And this is still dependency for many applications, for instance, collectd - system monitor daemon.

json/test_yajl.c
nuald: The increased memory consumption is expected. Streaming reading is out of scope for those particular tests - they've asked for that, but that would make the results table way too big.

Plus the streaming results could be confusing. For example, one of the most efficient way to do "streaming" reading is using `mmap`, but it's way too unrealistic for the majority of use-cases (as usually JSON is used to transfer data). Using network streams is closer to the production use-cases, but makes tests way too complicated. So, the common ground now is just parse buffers.

json/test_yajl.c
```diff
@@ -0,0 +1,211 @@
...
+		exit(EXIT_FAILURE);
+	}
+
+	yajl_parse(yajl, buf, file_stat.st_size);
+	yajl_complete_parse(yajl);
+
+	notify("stop");
+
+	yajl_free(yajl);
+	free(buf);
```
nuald, line 207: As we support GCC and Clang only, you may want to use `cleanup` attribute to avoid the issue like here, when the program exited above, but didn't free memory prior exiting. It's not that critical though, so it's up to you.

Some examples are here: https://www.berrange.com/posts/2020/01/31/libvirt-use-of-gcc-clang-extension-for-automatic-cleanup-functions/

 The same could be applied to `yajl_free`, and `close`.

json/test_yajl.c
```diff
@@ -0,0 +1,211 @@
...
+		exit(EXIT_FAILURE);
+	}
+
+	ret = fstat(fd, &file_stat);
+	if (ret < 0) {
+		fprintf(stderr, "Cannot fstat /tmp/1.json: %s\n", strerror(errno));
+		exit(EXIT_FAILURE);
+	}
+
+	buf = calloc(file_stat.st_size, sizeof(unsigned char));
```
nuald, line 184: `calloc` could be quite overhead here as if `read` didn't fill all the bytes, then the program exits anyway.

nuald (commented):
All minor, the only blocker is moving `read` before notification.

json/test_yajl.c
```diff
@@ -0,0 +1,211 @@
...
+
+	yajl_handle yajl = yajl_alloc(&ycallbacks, NULL, &ctx);
+
+	yajl_parse(yajl, (const unsigned char*)buf, size);
+	yajl_complete_parse(yajl);
+
+	yajl_free(yajl);
+
+	if (ctx.value.x != expected->x || ctx.value.y != expected->y || ctx.value.z != expected->z) {
+		fprintf(stderr, "{%lf, %lf, %lf} != {%lf, %lf, %lf}\n", ctx.value.x, ctx.value.y, ctx.value.z, expected->x, expected->y, expected->z);
```
nuald, line 142: A little bit code duplication here. Other tests use their own dump function for that, I guess it's possible to use it too:

```c
void dump(char *buf, coordinate_t coord);

char str_x[50];
dump(str_x, ctx.value);

char str_y[50];
dump(str_y, expected);

fprintf(stderr, "%s != %s\n", str_x, str_y);
```

json/test_yajl.c
```diff
@@ -0,0 +1,211 @@
...
+
+	notify_with_pid("C/" COMPILER " (yajl)");
+
+	ret = read(fd, buf, file_stat.st_size);
+	if (ret != file_stat.st_size) {
+		fprintf(stderr, "Cannot read /tmp/1.json: %s\n", strerror(errno));
+		exit(EXIT_FAILURE);
+	}
+
+	yajl_parse(yajl, buf, file_stat.st_size);
```
nuald, line 201: A little bit abstraction leak here (file size), and I guess it would be nice to separate file operations and parsing by introduce an auxiliary struct:

```c
struct str_s {
  char *buf;
  size_t buf_sz;
};
typedef struct str_s str_t;

void str_free(str_t* s) { free(s->buf); }

str_t readfile(const char* filename) {...}

int main() {
  ...
  __attribute__((__cleanup__(str_free))) str_t str = readfile("/tmp/1.json");
  ...
  yajl_parse(yajl, str.buf, str.buf_sz);
}

```

nuald (commented):
Somehow Github lost my comments in the last review, re-submitting them. Minor, it's up to you, but it would be more consistent with other tests.

json/test_yajl.c
```diff
@@ -0,0 +1,184 @@
...
+	if (ctx->depth == 1 && ctx->in_coord) {
+		ctx->c.x /= ctx->len;
+		ctx->c.y /= ctx->len;
+		ctx->c.z /= ctx->len;
+	}
+
+	return 1;
+}
+
+static yajl_callbacks ycallbacks = {
```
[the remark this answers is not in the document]
matwey: Done

nuald (approved):

PR #456, opened 2023-03-08
Rust improvements

ricvelozo:
New macro `notify!` to simplify use (exported as a crate to be shared), and other minor improvements.


nuald:
Looks good. Please note that originally the code duplication (including for `notify`) was intentional so the code could be copy pasted into other projects. However, considering that `notify` functionality has no use for other projects, it's totally fine to extract it into the common module (like it's done for C++).

Issue #457, opened 2023-03-09
Assertions


ricvelozo:
I want to use assertions in `verify()` functions to simplify the code. Does the error output need to be exact? Example:

```rust
if left != right {
    eprintln!("{left} != {right}");
    process::exit(-1);
}
````

Will become just:

```rust
assert_eq!(left, right);
````

The error output:

```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `80`,
 right: `82`', src/main.rs:2:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

ricvelozo:
And the `verify()` function needs to be called at runtime? Because in Swift we must disable the `-Ounchecked` flag to the assertion don't be optimized away.

nuald:
The code is not tested in debug mode at all, so whatever verification is used, should work in the release mode too.

ricvelozo:
> The code is not tested in debug mode at all, so whatever verification is used, should work in the release mode too.

Alright, I will use `fatalError()` in Swift (not optimized away). But the `-Ounchecked` flag is another one to go to the #290 discussion.

PR #458, opened 2023-03-09
Use assertions in Rust code

ricvelozo:
See #457

### DIFF
--- a/matmul/matmul.rs/src/main.rs
+++ b/matmul/matmul.rs/src/main.rs
@@ -58,10 +58,7 @@ fn main() {
 
     let left = calc(101);
     let right = -18.67;
-    if (left - right).abs() > 0.1 {
-        eprintln!("{left} != {right}");
-        process::exit(-1);
-    }
+    assert!((left - right).abs() <= 0.1);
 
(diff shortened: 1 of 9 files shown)


matmul/matmul.rs/src/main.rs
```diff
@@ -58,10 +58,7 @@ fn main() {
 
     let left = calc(101);
     let right = -18.67;
-    if (left - right).abs() > 0.1 {
-        eprintln!("{left} != {right}");
-        process::exit(-1);
-    }
+    assert!((left - right).abs() <= 0.1);
```
nuald, line 61: I'd prefer that the exact values would be shown here, e.g.

```rs
assert!((left - right).abs() <= 0.1, "{left} != {right}");
```

The same for other float point comparisons. Using the special macro is fine too, like mentioned in  https://stackoverflow.com/questions/30856285/assert-eq-with-floating-point-numbers-and-delta

ricvelozo:
I disabled the `-Ounchecked` flag anyway because it was giving worse performance with the wrapping arithmetic I using now.

PR #461, opened 2023-05-27
Add staged Scala variant

philnguyen:
This one starts from the main Scala variant, adops Scala 3's preferred [quiet  syntax](https://docs.scala-lang.org/scala3/reference/other-new-features/control-syntax.html), then uses Scala's [multi-stage programming](https://docs.scala-lang.org/scala3/reference/metaprogramming/staging.html) to apply the same technique in the `Racket (Syntax Objects)` variant. And the JVM is wicked fast!!

Unfortunately I couldn't figure out the more lightweight way to get it to work with just the compiler directly (i.e. `scalac -with-compiler -d out Main.scala` as they suggest in their docs), so I have to use [sbt](https://docs.scala-lang.org/getting-started/sbt-track/getting-started-with-scala-and-sbt-on-the-command-line.html).

The commands I use to build and run are currently as follow:
```
$ cd scala-staged
$ sbt "run ../bench.b"
```

### DIFF
--- a/brainfuck/scala-staged/build.sbt
+++ b/brainfuck/scala-staged/build.sbt
@@ -1,0 +1,14 @@
+val scala3Version = "3.2.2"
+
+lazy val root = project
+  .in(file("."))
+  .settings(
+    name := "bf",
+    version := "0.1.0",
+
+    scalaVersion := scala3Version,
+
+    libraryDependencies ++= Seq(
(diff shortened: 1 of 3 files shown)


nuald:
The test has been embedded into the suite in #472 . Required minor modifications to make it compatible with the test suite, but it's only about main object, logic hasn't been changed.

PR #479, opened 2023-11-09
Reorganize Go code

### LINKED ISSUE #478: Reorganize and improve go code
When working on #476 I noticed that Go benchmarks are a tough to work on because they are all in the same directory so editors interpret that as one big package with a lot of duplicate function names.

I propose changing the layout of the Go benchmarks to be more like the Rust layout. There would be a `common/go` package with a `Run` function that runs the benchmarked function and sends the results to `:9001`.

Each test directory would have a `test-go` subdirectory that contains a separate package for each Go benchmark. For example there would be `json/json-go/goccy` and `json/json-go/jsoniter`.

If this gets a 👍 I can do the work. I did a little bit of exploratory work with a few json benchmarks, and it won't take too long to apply that to the rest of the Go benchmarks.

WillAbides:
closes #478

- `common/go` is a new Go package that contains `common.RunBenchmark`, which
  is a common benchmarking function that can be used by all the go
  benchmarks.
- `common.RunBenchmark` now translates from the compiler name to the final name.
  e.g. `gc` becomes `Go` and `gccgo` becomes `Go/gccgo`.
- base64x now reports as "Go (base64x)" instead of just "Go". Previously there
  were two "Go" entries for base64.
- Each test case directory now has a `go` subdirectory that contains a
  `Run` function at the top level and subdirectories for each go implementation.
- Errors are no longer ignored. Go optimizes err checks pretty well, so it
  won't affect performance by enough to show up on the millisecond scale
  reporting. However it is more idiomatic Go code.
- `common.RunBenchmark` creates tcp connections before running the benchmark
  so that the time to create the connections is not included in the reported
  time.
- gccgo is now compiled using `go build -compiler gccgo` instead of directly
  calling gccgo. This is more convenient when working with modules.
- Makefiles are refactored to use the new patterns which allowed me to
  deduplicate a good chunk of it and move it to commands.mk


nuald:
I've tried it locally and got:
```
go build -C go/base64/ -ldflags="-s -w" -o /src/base64/go/base64/target/benchmark .
error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.
make: *** [../common/commands.mk:109: go/base64/target/benchmark] Error 1
```

Could you please fix it?

WillAbides:
I added `-buildvcs=false` to the go build command which _should_ fix that, but I couldn't make it happen locally so I'm not sure about the fix either.

nuald:
It seems the only issue I had, I've tested the branch, and it all works fine, thanks!

PR #486, opened 2024-01-26
php boost

sibkod:
Removed the creation of a local variable.
This gave an almost 10% increase in speed


nuald:
That's a micro-optimization, and not the code that average software developer would write. Another story is using `foreach` though  - that's seems more appropriate and could have the same effect.

sibkod:
Using foreach is what the average developer will write, but it will be even slower than it is now

PR #487, opened 2024-05-22
Some updates to Rust code

ricvelozo:
* Use slices and tuples in base64 where possible instead of Vec
* More generic code and better variables names in bf.rs

### DIFF
--- a/base64/base64.rs/src/bin/base64.rs
+++ b/base64/base64.rs/src/bin/base64.rs
@@ -1,5 +1,4 @@
 use base64::Engine;
-use std::io::Write;
 use std::time::Instant;
 use std::{process, str};
 use utils::notify;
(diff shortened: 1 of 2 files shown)


base64/base64.rs/src/bin/base64.rs
```diff
@@ -10,17 +9,27 @@ const TRIES: usize = 8192;
 fn main() {
     let base64_engine = base64::engine::general_purpose::STANDARD;
 
-    for [src, dst] in &[["hello", "aGVsbG8="], ["world", "d29ybGQ="]] {
-        let encoded = base64_engine.encode(src);
+    for (src, dst) in [(b"hello", b"aGVsbG8="), (b"world", b"d29ybGQ=")] {
+        let mut encoded = [0u8; 8];
+        base64_engine.encode_slice(src, &mut encoded).unwrap();
```
nuald, line 14: It's fine to extend the tests, but the original idea is testing the API used in the benchmark as we don't have the visibility into the hot path execution (we already had the situation when API didn't work correctly). Please add (or revert) the testing of `encode`/`decode` functions used in the benchmark.

nuald (commented):
Just a minor comment. Also, I'd like to note that `decode`/`encode` are intentionally used as memory allocation is a part of the test. The only additional optimization allowed (as it's done in C) - pre-allocate a chunk on memory in stack just before the operation inside the loops and put the result into that chunk. While it's unfair for GC languages, it's idiomatic for native languages to use stack for data.

PR #499, opened 2024-11-19
Remove obsolete increments of len variable in Odin test

ttytm:
Adding the additional len variable and mutating it in the loop introduces a slowdown of about 5-10%.

The formatting changes are made by ols/odinfmt.


nuald:
I've tested, but I don't see any differences. Please note that the tests run on the server (Xeon) CPU with bigger caches, extended instructions and branch predictors, so this kind of optimization could be unnecessary.  For your reference, I've attached the flamegraph and you could see that the most CPU time is spending in the actual parser.
odin

PR #502, opened 2024-11-20
Add JSON C3

ttytm:
I'd add an bfk bench in a separate PR if an initial working bench for the language has been added.

### DIFF
--- a/common/commands.mk
+++ b/common/commands.mk
@@ -9,6 +9,7 @@ V_FLAGS := -prod -no-bounds-checking -prealloc -cflags "-Wno-error=implicit-func
 V_VSL_CBLAS_FLAGS := $(V_FLAGS) -d cblas
 ZIG_FLAGS := -O ReleaseFast
 
+C3_BUILD =	c3c -05 --safe=no $^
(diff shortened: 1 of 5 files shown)


common/commands.mk
```diff
@@ -9,6 +9,7 @@ V_FLAGS := -prod -no-bounds-checking -prealloc -cflags "-Wno-error=implicit-func
 V_VSL_CBLAS_FLAGS := $(V_FLAGS) -d cblas
 ZIG_FLAGS := -O ReleaseFast
 
+C3_BUILD =	c3c -05 --safe=no $^
```
ttytm, line 12: Regarding the --safe flag: it's to disable bounds check, similar as it's done for other languages.

Ref.:
```
--safe=<yes|no>            - Turn safety (contracts, runtime bounds checking, null pointer checks etc) on or off.
```

nuald: Yes, that's expected to be set, thanks for the clarification.

nuald:
Follow-up has published with the results and few minor issues fixes.

ttytm:
> Follow-up has published with the results and few minor issues fixes.

Very nice thank you!

